PR #2, opened 2015-07-07
Added CORS support

om-sharma:
This commit adds the capability of CORS. Currently it supports communication with all the external domains also it allows all the http methods. We can discuss more in details about it and can limit it off if required.

### DIFF
--- a/src/test/java/uk/gov/register/presentation/functional/FunctionalTest.java
+++ b/src/test/java/uk/gov/register/presentation/functional/FunctionalTest.java
@@ -1,5 +1,7 @@
 package uk.gov.register.presentation.functional;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.net.HttpHeaders;
 import io.dropwizard.client.JerseyClientBuilder;
 import io.dropwizard.testing.ConfigOverride;
 import io.dropwizard.testing.ResourceHelpers;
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/register/presentation/functional/FunctionalTest.java
```diff
@@ -41,6 +45,26 @@ public void shouldConsumeMessageFromKafkaAndShowAsLatest() throws Exception {
 
     }
 
+    @Test
+    public void appSupportsCORS() {
```
philandstuff, line 49: is it possible to test this in a [Resource test](https://dropwizard.github.io/dropwizard/manual/testing.html#testing-resources) rather than a functional test? Functional tests are expensive and this behaviour isn't a property of the whole app, just of this one resource.

philandstuff, line 49: @om-sharma pointed out to me in slack that this won't work as a resource test as the functionality is provided by a Filter instead.  The full DropwizardAppRule test is the only way to test this functionality.  Sadly that means you need to connect to a real database and kafka server.

I'd rather have very few functional tests and focus them on ensuring that the whole app is glued together correctly.  The CrossOriginFilter is part of jetty so I think we can be quite confident that it will work.

om-sharma, line 49: The test confirms that app provides the CORS support. It has to be a functional test, doesn't matter  how it is implemented. I am agree that functional test suite should provide quick feedback regardless of number of tests in it.

PR #3, opened 2015-07-07
Added CORS support

om-sharma:
This commits adds the capability of CORS. Currently it supports communication with all the external domains also it allows all the http methods. We can discuss more in details about it and can limit it off if required.


philandstuff:
:+1:

PR #4, opened 2015-07-07
Show multiple entries in /latest.json

philandstuff:
PR #1 started displaying JSON properly at /latest.json.  This PR ensures
that we can present more than one entry.

To do this, I needed to update FunctionalTest to clean the database each
time.  Because @BeforeClass methods run after TestRule setup, and we
needed to clean the database before starting the app, I needed to create
a new TestRule for cleaning the database.

Because the CleanDatabaseRule needs the db connection url, I've pulled
it out of the yaml file into the FunctionalTest class.


philandstuff:
oh I should have said as well -- this displays entries in reverse order (latest first), copying the alpha product's behaviour.

PR #5, opened 2015-07-08
Added register home page

om-sharma:
Copied home page from alpha register app, also copied required css and vendor assets also.


philandstuff:
Looks good.  Can you hardcode some specific (made-up) values rather than cluttering the page with `<-- placeholder -->` stuff?

Also is it worth splitting PresentationResource into HomeResource and LatestResource as they're pretty unrelated right now?

PR #11, opened 2015-07-20
Escaping csv and tsv entries

om-sharma:
Refactored the CsvEntity class in two classes, one for Csv representation and second for tsv representation. Earlier implementation was not escaping any data.

### DIFF
--- a/src/test/java/uk/gov/register/presentation/entity/CsvEntityTest.java
+++ b/src/test/java/uk/gov/register/presentation/entity/CsvEntityTest.java
@@ -31,7 +31,7 @@ public class CsvEntityTest {
         AbstractView view = new SingleResultView("foo", node);
         Object result = csvEntity.convert(view);
 
-        assertThat(result.toString(), equalTo("hash,key1,key2,key3,key4\nhash1,valu\te1,\"val,ue2\",\"val\"\"ue3\",\"val\nue4\""));
+        assertThat(result.toString(), equalTo("hash,key1,key2,key3,key4\r\nhash1,valu\te1,\"val,ue2\",\"val\"\"ue3\",\"val\nue4\""));
     }
 
 }
--- a/src/test/java/uk/gov/register/presentation/entity/CsvEntityTest.java
+++ b/src/test/java/uk/gov/register/presentation/entity/CsvEntityTest.java
@@ -1,0 +1,37 @@
+package uk.gov.register.presentation.entity;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.collect.ImmutableMap;
+import org.junit.Test;
+import uk.gov.register.presentation.mapper.JsonObjectMapper;
+import uk.gov.register.presentation.view.AbstractView;
+import uk.gov.register.presentation.view.SingleResultView;
+
+import java.util.Map;
+
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.junit.Assert.assertThat;
+
+public class CsvEntityTest {
+    @Test
+    public void csv_entriesAreEscaped() {
+        CsvEntity csvEntity = new CsvEntity();
+
+        Map jsonMap = ImmutableMap.of(
+                "hash", "hash1",
+                "entry", ImmutableMap.of(
+                        "key1", "valu\te1",
+                        "key2", "val,ue2",
+                        "key3", "val\"ue3",
+                        "key4", "val\nue4"
+                )
+        );
+        JsonNode node = JsonObjectMapper.convert(jsonMap, JsonNode.class);
+
+        AbstractView view = new SingleResultView("foo", node);
+        Object result = csvEntity.convert(view);
+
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/register/presentation/entity/CsvEntityTest.java
```diff
@@ -0,0 +1,37 @@
...
+                        "key3", "val\"ue3",
+                        "key4", "val\nue4"
+                )
+        );
+        JsonNode node = JsonObjectMapper.convert(jsonMap, JsonNode.class);
+
+        AbstractView view = new SingleResultView("foo", node);
+        Object result = csvEntity.convert(view);
+
+        assertThat(result.toString(), equalTo("hash,key1,key2,key3,key4\nhash1,valu\te1,\"val,ue2\",\"val\"\"ue3\",\"val\nue4\""));
```
philandstuff: this doesn't look right. Are you sure you want a literal newline character here, rather than the characters `\` and `n`?

src/test/java/uk/gov/register/presentation/entity/TsvEntityTest.java
```diff
@@ -0,0 +1,38 @@
...
+                        "key3", "val\"ue3",
+                        "key4", "val\nue4"
+                )
+        );
+        JsonNode node = JsonObjectMapper.convert(jsonMap, JsonNode.class);
+
+        AbstractView view = new SingleResultView("someTemplate", node);
+        Object result = csvEntity.convert(view);
+
+        assertThat(result.toString(), equalTo("hash\tkey1\tkey2\tkey3\tkey4\nhash1\tvalue1\t\"val\tue2\"\t\"val\"\"ue3\"\t\"val\nue4\""));
```
philandstuff: I'm a little uneasy about supporting newlines in data in tsv at all -- one of the benefits of tsv is it fits tools such as cut(1) and awk(1); but this breaks when your data has newlines.

src/test/java/uk/gov/register/presentation/entity/CsvEntityTest.java
philandstuff: also, csv lines should be separated with `\r\n` not `\n` -- see [rfc 4180](https://tools.ietf.org/html/rfc4180#section-2)

om-sharma: 1. I specifically want to test that if content contains newline '\n' character it is escaped. I am using apache commons StringEscapeUtils.escapeCsv() method which does the escaping.
2. Yes every row should be separated by CRLF. Let me change it.

src/test/java/uk/gov/register/presentation/entity/TsvEntityTest.java
om-sharma: Same as it CSV, tsv escaping is again commons lang. it is just the test, if content contains '\n' it will be escaped. If we don't want '\n' in content we have to add handling where we input the data.
The other option is that we replace '\n', '\r' with empty string if data content contains them.

src/test/java/uk/gov/register/presentation/entity/CsvEntityTest.java
philandstuff: if it were escaped, I'd expect to see `"\"val\\nue4\""` -- note the double backslash.  Instead, you have a literal newline character which has been merely quoted, not escaped.

philandstuff: mind you, the CSV RFC doesn't say anything about escaping newlines, just quoting them.  :cry:

src/test/java/uk/gov/register/presentation/entity/TsvEntityTest.java
philandstuff: hmm, shrug. @psd any thoughts?

src/test/java/uk/gov/register/presentation/entity/CsvEntityTest.java
om-sharma: I was following 'org.apache.commons.lang3.StringEscapeUtils.escapeCsv()'. Didn't read RFC, let me check that and confirm why apache escape method is doing differently.

philandstuff: no you misunderstand.  commons-lang conforms with the RFC.  it's just annoying that the RFC uses literal newlines :(

philandstuff: @om-sharma and I checked this together and it opens in libreoffice so this is fine.

src/test/java/uk/gov/register/presentation/entity/TsvEntityTest.java
philandstuff: let's go with this for now, and @psd will define some test cases in the upcoming reference register card.

PR #12, opened 2015-07-20
return 404 for non-primary-key path

philandstuff:
We have defined resources for `/hash/<value>` and `/<primary-key>/<value>`.
Any other path is not defined.  The http way to respond that a path is
not recognized is to say 404 Not Found.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/resource/SearchResource.java
+++ b/src/main/java/uk/gov/register/presentation/resource/SearchResource.java
@@ -42,7 +42,7 @@ public class SearchResource extends ResourceBase {
             return buildResponse(new SingleResultView("/templates/entry.mustache", entry.isPresent() ? entry.get() : null));
         }
 
-        throw new BadRequestException("Key: " + key + " is not primary key of the register.");
+        throw new NotFoundException();
     }
 
     @GET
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/register/presentation/resource/SearchResource.java
```diff
@@ -42,7 +42,7 @@ public Response findByPrimaryKey(@PathParam("primaryKey") String key, @PathParam
             return buildResponse(new SingleResultView("/templates/entry.mustache", entry.isPresent() ? entry.get() : null));
         }
 
-        throw new BadRequestException("Key: " + key + " is not primary key of the register.");
+        throw new NotFoundException();
```
om-sharma, line 45: In my opinion it should be 400. For a particular register we don't allow search with any key other than primary key. If you try to search, it is bad request because you typed wrong key. while 404 in this particular case means that the 'key' is correct but there is no resource found with the value.

philandstuff, line 45: I disagree. We don't define a resource `/{key}/{value}` and require that `{key}` = `postcode`, we define a resource `/postcode/{value}` and leave everything else undefined.

Think of it from the point of view of the consumer, who knows nothing about the back-end implementation.  They know that `school.openregister.org/hash/{value}` is defined, and that `school.openregister.org/school/{value}` is defined, but there's nothing documented leading them to expect anything of `school.openregister.org/{anythingelse}/{value}`, so it's natural for it to be a 404.

PR #17, opened 2015-07-22
Turtle representation

om-sharma:
Added functionality for all api methods to support turtle representation. This commit also include the refactoring to use Entry object to render json, also all writers are using Entry object now and view provides other methods needed for only html.


src/main/java/uk/gov/register/presentation/dao/RecentEntryIndexQueryDAO.java
```diff
@@ -8,27 +7,24 @@
 import org.skife.jdbi.v2.sqlobject.customizers.SingleValueResult;
 import uk.gov.register.presentation.Entry;
 import uk.gov.register.presentation.mapper.EntryMapper;
-import uk.gov.register.presentation.mapper.JsonNodeMapper;
 
 import java.util.List;
 
-@RegisterMapper(JsonNodeMapper.class)
```
philandstuff, line 15: you can delete JsonNodeMapper too now

om-sharma, line 15: Deleted JsonNodeMapper class in next commit. :)

src/main/java/uk/gov/register/presentation/mapper/JsonNodeMapper.java
```diff
@@ -1,15 +0,0 @@
...
-import org.skife.jdbi.v2.StatementContext;
-import org.skife.jdbi.v2.tweak.ResultSetMapper;
-
-import java.sql.ResultSet;
-import java.sql.SQLException;
-
-public class JsonNodeMapper implements ResultSetMapper<JsonNode> {
-    @Override
-    public JsonNode map(int index, ResultSet r, StatementContext ctx) throws SQLException {
-        return JsonObjectMapper.convertToJsonNode(r.getBytes("entry"));
```
philandstuff, line 13: JsonObjectMapper.convertToJsonNode is now unused :) :fire:

src/main/java/uk/gov/register/presentation/representations/TurtleWriter.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Produces(ExtraMediaType.TEXT_TTL)
+public class TurtleWriter extends RepresentationWriter<AbstractView> {
+
+
+    @Override
+    public boolean isWriteable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return type.isAssignableFrom(SingleResultView.class) || type.isAssignableFrom(ListResultView.class);
```
philandstuff: this is backwards. you want `SingleResultView.class.isAssignableFrom(type)` to ensure that `type` is a subtype of SingleResultView

src/main/java/uk/gov/register/presentation/representations/TurtleWriter.java
```diff
@@ -0,0 +1,83 @@
...
+            entries.add((Entry) o);
+            return entries;
+        }
+    }
+
+    //TODO: this should be retrieved from register call
+    private static final String registerBaseUri = "http://localhost:9000";
+
+    @SuppressWarnings("unchecked")
+    private List entryFields(Map map) {
```
philandstuff: you can remove the warning suppression here with the signature `private List<String> entryFields(Map<String,?> map)`

src/main/java/uk/gov/register/presentation/representations/TurtleWriter.java
```diff
@@ -0,0 +1,83 @@
...
+
+        List<String> fields = entryFields(JsonObjectMapper.convert(entries.get(0).getContent(), Map.class));
+        for (Entry entry : entries) {
+            writeRow(entry, fields, entityStream);
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private List<Entry> getEntries(Object o) {
+        if (o instanceof List) {
```
philandstuff: Would it be nicer to do the instanceof checks on `view` rather than on the result of `view.get()`?

src/main/java/uk/gov/register/presentation/representations/TurtleWriter.java
```diff
@@ -0,0 +1,83 @@
...
+    //TODO: this should be retrieved from register call
+    private static final String registerBaseUri = "http://localhost:9000";
+
+    @SuppressWarnings("unchecked")
+    private List entryFields(Map map) {
+        return new ArrayList<>((map.keySet()));
+    }
+
+    public void writeRow(Entry entry, List<String> fields, OutputStream entityStream) {
+        try {
```
philandstuff: you don't need this try block if you just declare `throws IOException` here. also does this method need to be public?

PR #20, opened 2015-07-24
Supporting multiple register by one app instance

om-sharma:
Changes confirm that an app instance can support multiple registers. Every register connects to it's own named database.
Also removed kafka integration so that presentation app is pure read view api. Presentation app now simply connect to backend (currently postgres) and provides the views, It does not writes in backend.


philandstuff:
so if presentation doesn't consume kafka, what does? :confused:

philandstuff:
after conversation in slack: closing this PR as it breaks updates from kafka to postgres.

We're trying to set up an end-to-end environment and don't want to break that for the moment.

The answer may be to switch postgres out for elasticsearch and use its kafka input plugin, at which point this PR might be redundant.

PR #21, opened 2015-07-27
Supporting multiple register by one app instance

om-sharma:
Changes confirm that an app instance can support multiple registers. Every register connects to it's own named database. 
For every register a separate kafka consumer starts which connects with the kafka topic with register name.


om-sharma:
Closing the pull request, so that it doesn't run the build everytime for commit. needs to merge master and redo some functionality.

PR #25, opened 2015-08-03
Code deploy integration

om-sharma:
These changes enables app to deploy master to aws environment using code deploy.


deploy/scripts/stop-service.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash -x
-RES=`ps aux | grep -v grep | grep 'PresentationApplication' | awk '{print $2}'`
+RES=`ps aux | grep -v grep | grep 'presentation.jar' | awk '{print $2}'`
 if [[ ! -z "$RES" ]]; then
     kill $RES
```
philandstuff, line 4: TODO: this should be `docker stop`.  Also there should be a `validate.sh` that pings the [dropwizard health check endpoint](https://dropwizard.github.io/dropwizard/manual/core.html#health-checks)

PR #26, opened 2015-08-04
Using prefix 'field' for all fields in turtle representation.

om-sharma:
Using prefix 'field' for all fields in turtle representation.


philandstuff:
doesn't validate with http://rhizomik.net/html/redefer/rdf2svg-form/

om-sharma:
Removed the trailing ';' from subject part.

PR #29, opened 2015-08-10
Validate script failure

om-sharma:
using while loop to validate that the application started properly. it retries every 5 seconds. validation step timeout is 5 minutes. Ideally it should come up very quick like 10-15 seconds but once it took 3 minutes so kept the 5 minute timeout.


philandstuff:
looks good, waiting for travis before merge

PR #34, opened 2015-08-19
Thymeleaf template support

om-sharma:
This change enables the use of thymeleaf templates. Removed the mustache renderer and not using mustache templates now.


om-sharma:
closing it. need to delete some commented code.

PR #35, opened 2015-08-19
Thymeleaf template support

om-sharma:
This change enables the use of thymeleaf templates. Removed the mustache renderer and not using mustache templates now.


src/main/java/thymeleaf/ThymeleafContext.java
```diff
@@ -0,0 +1,49 @@
+package thymeleaf;
```
philandstuff: is this the right package?

src/main/java/thymeleaf/ThymeleafViewRenderer.java
```diff
@@ -0,0 +1,63 @@
...
+        templateResolver.setPrefix(prefix);
+        templateResolver.setCacheable(cacheable);
+        engine = new TemplateEngine();
+        engine.setTemplateResolver(templateResolver);
+
+    }
+
+    @Override
+    public boolean isRenderable(View view) {
+        return view.getTemplateName().endsWith(suffix);
```
philandstuff: shouldn't this check for `instanceof ThymeleafView` instead of doing a string comparison?

src/main/java/uk/gov/register/presentation/app/PresentationApplication.java
```diff
@@ -67,11 +65,15 @@ public void run(PresentationConfiguration configuration, Environment environment
...
 
         DropwizardResourceConfig resourceConfig = environment.jersey().getResourceConfig();
-        resourceConfig.property(ServerProperties.MEDIA_TYPE_MAPPINGS, ImmutableMap.of(
+
+        ImmutableMap<String, MediaType> representations = ImmutableMap.of(
                 "csv", ExtraMediaType.TEXT_CSV_TYPE,
                 "tsv", ExtraMediaType.TEXT_TSV_TYPE,
                 "ttl", ExtraMediaType.TEXT_TTL_TYPE,
-                "json", MediaType.APPLICATION_JSON_TYPE));
+                "html", MediaType.TEXT_HTML_TYPE,
```
philandstuff: to be removed (discussed IRL -- this is a separate concern from thymeleaf support)

philandstuff:
:+1: finished reviewing, please address inline comments

src/main/java/thymeleaf/ThymeleafViewRenderer.java
om-sharma: It should be string ends with check to confirm that the ThymeleafView object is created with the html template whose name ends with .html.

src/main/java/uk/gov/register/presentation/app/PresentationApplication.java
om-sharma: done!

src/main/java/thymeleaf/ThymeleafContext.java
om-sharma: moved under uk.gov.register package.

PR #36, opened 2015-08-19
404 500 pages

om-sharma:
Added 404 and 500 pages. 404.html is rendered for NotFoundException while 500.html is rendered for any other exception caught.


om-sharma:
Will reopen the request. seems like some issue with the commit.

PR #37, opened 2015-08-19
Removed hardcoded register base url from turtle representation.

om-sharma:
Removed hardcoded register base url from turtle representation.
Small refactoring to make sure that TurtleWriter extends specific generic type of RepresentationWriter which is ThymeleafView.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/representations/TurtleWriter.java
+++ b/src/main/java/uk/gov/register/presentation/representations/TurtleWriter.java
@@ -4,8 +4,10 @@ import io.dropwizard.views.View;
 import uk.gov.register.presentation.Record;
 import uk.gov.register.presentation.resource.ResourceBase;
 
+import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/register/presentation/representations/TurtleWriter.java
```diff
@@ -14,12 +17,17 @@
...
-import java.util.*;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 @Produces(ExtraMediaType.TEXT_TTL)
-public class TurtleWriter extends RepresentationWriter<View> {
-    private static final String PREFIX ="@prefix field: <http://field.openregister.org/field/>.\n\n";
+public class TurtleWriter extends RepresentationWriter<ThymeleafView> {
```
philandstuff: this doesn't need to be specific to `ThymeleafView`

PR #39, opened 2015-08-19
Added 404 and 500 pages.

om-sharma:
Added 404 and 500 pages. 404.html is rendered for NotFoundException while 500.html is rendered for any other exception caught.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,3 +1,6 @@
+server:
+  registerDefaultExceptionMappers: false
+
 database:
   driverClass: org.postgresql.Driver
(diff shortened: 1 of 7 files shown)


philandstuff:
For some reason, if I cause a 500 error on this branch, I get a JSON response, not HTML

philandstuff:
It turns out this is because [dropwizard registers its own `ExceptionMapper<Throwable>`](https://github.com/dropwizard/dropwizard/blob/v0.8.2/dropwizard-core/src/main/java/io/dropwizard/server/AbstractServerFactory.java#L480-L481)

philandstuff:
see also dropwizard/dropwizard#710

PR #40, opened 2015-08-20
Support csv and tsv for single entry pages

philandstuff:
The template for a single entry page has links to csv and tsv
representations, but we had no support for rendering them.  We could
either remove the links, or support these representations for single
entries.  This commit adds the required support to tsv and csv.


philandstuff:
@psd you finally got your way :)

PR #42, opened 2015-08-28
Fast template feedback

om-sharma:
Enables hot loading templates in dev environment.
 thymeleaf renderer uses FileResourceLoader when a system property 'baseDirForTemplates' is set. it enables application to load any changes in the templates while server is running.
 added shell scripts run-application.sh to start application. it invokes gradle task run to start the application.


philandstuff:
can you update the README?

om-sharma:
Update README.

PR #43, opened 2015-09-02
ensure presentation postgres role exists

philandstuff:
We couldn't get this working on @stephenjoe1's machine without this
user.


om-sharma:
We can enforce everyone to has 'postgres' user to start the application server in dev mode. To fix the problem change config.yaml to specify default user as 'postgres'.

PR #48, opened 2015-09-02
Using postgres as default user. 

om-sharma:
Using postgres as default user. to run tests and also to start application server so that no conflicts.
 start server failed at stephen's machine because run application uses config.yaml which has default user as presentation.


philandstuff:
(for posterity, this is a better fix than #43)

PR #50, opened 2015-09-04
Linking fields on entry and entries page to alpha field register.

om-sharma:
Linking fields on entry and entries page to alpha field register.

### DIFF
--- a/src/main/resources/templates/entry.html
+++ b/src/main/resources/templates/entry.html
@@ -8,7 +8,9 @@
     <div class="small-12 large-8 columns">
         <h1 class="entry-header" th:text="${record.entry.get('school')}">Entry Name Header</h1>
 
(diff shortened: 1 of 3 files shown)


src/main/resources/templates/entry.html
```diff
@@ -8,7 +8,9 @@
     <div class="small-12 large-8 columns">
         <h1 class="entry-header" th:text="${record.entry.get('school')}">Entry Name Header</h1>
 
-        <p class="entry-source">An entry in the <a href="/" th:text="${@uk.gov.register.presentation.RegisterNameExtractor@capitalizedRegisterName(#httpServletRequest.getHeader('Host')) + ' register'}"></a></p>
+        <p class="entry-source">An entry in the <a href="/"
```
philandstuff, line 11: what's with all the whitespace here?

philandstuff, line 11: oh wait, I think I got confused  by github's formatting

PR #51, opened 2015-09-04
actually log exceptions when they are thrown

philandstuff:
Currently, if we 500 there's no useful way to work out what actually happened.


davidillsley:
Looks good.

philandstuff:
ta for the review @davidillsley

PR #54, opened 2015-09-10
Created separate RecordView object for html view.

om-sharma:
Renamed Record to DbRecord to keep this object only for to deserialize data from database.
 Did the related refactoring to achieve above.


src/main/java/uk/gov/register/presentation/RecordConverter.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public RecordView convert(DbRecord dbRecord) {
+        Iterable<Map.Entry<String, JsonNode>> fields = () -> dbRecord.getEntry().fields();
+        Stream<Map.Entry<String, JsonNode>> fieldStream = StreamSupport.stream(fields.spliterator(), false);
+        return new RecordView(dbRecord.getHash(), requestContext.getRegisterPrimaryKey(), fieldStream.collect(Collectors.toMap(Map.Entry::getKey, this::convert)));
+    }
+
+
+    private FieldValue convert(Map.Entry<String, JsonNode> entry) {
+        return fieldsConfiguration.getField(entry.getKey()).getRegister().isPresent() ? new LinkValue(entry.getKey(), entry.getValue().textValue()) : new StringValue(entry.getValue().textValue());
```
philandstuff: this assumes that field.getRegister() is the same string as the fieldName.  it might fail specifically for the `registry` field which links to the `public-body` register, not to the (non-existent) `registry` register

src/main/java/uk/gov/register/presentation/representations/FieldValueJsonSerializer.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.register.presentation.representations;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import uk.gov.register.presentation.FieldValue;
+
+import java.io.IOException;
+
+public class FieldValueJsonSerializer extends JsonSerializer<FieldValue> {
```
philandstuff: why do we need this class? what does it do?

om-sharma: this is not required, unused as such. will remove it in next commit.

src/main/java/uk/gov/register/presentation/RecordConverter.java
om-sharma: Correct! just committed now, the change in LinkValue takes register name as well as field name.

PR #55, opened 2015-09-10
Add YAML representation

philandstuff:
As part of this, I've made a RepresentationsTest which is parameterized over different representations; here, it tests both turtle and yaml.


src/main/java/uk/gov/register/presentation/representations/YamlWriter.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.MessageBodyWriter;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+@Produces(ExtraMediaType.TEXT_YAML)
+@Service
+public class YamlWriter implements MessageBodyWriter<View> {
```
om-sharma: This class should extend RepresentationWriter instead of implementing MessageBodyWriter.

philandstuff: I tried that, but RepresentationWriter assumes that a single element serializes the same as a list of one element, which isn't true in YAML :/

PR #56, opened 2015-09-11
Implemented history resource

om-sharma:
Added history page and a resource to return all the versions of an entry.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/resource/HistoryResource.java
+++ b/src/main/java/uk/gov/register/presentation/resource/HistoryResource.java
@@ -37,7 +37,7 @@ public class HistoryResource {
         throw new NotFoundException();
     }
 
-    @JsonSerialize(using = VersionJsonSerializer.class)
+    @JsonSerialize(using = ListVersionViewJsonSerializer.class)
     public static class ListVersionView extends ThymeleafView {
         private final List<Version> versions;
 
@@ -51,7 +51,7 @@ public class HistoryResource {
         }
     }
 
-    private static class VersionJsonSerializer extends JsonSerializer<ListVersionView> {
+    private static class ListVersionViewJsonSerializer extends JsonSerializer<ListVersionView> {
         @Override
         public void serialize(ListVersionView listVersion, JsonGenerator gen, SerializerProvider serializers) throws IOException {
             JsonSerializer<Object> listSerializer = serializers.findValueSerializer(List.class);
--- a/src/main/java/uk/gov/register/presentation/resource/HistoryResource.java
+++ b/src/main/java/uk/gov/register/presentation/resource/HistoryResource.java
@@ -1,0 +1,61 @@
+package uk.gov.register.presentation.resource;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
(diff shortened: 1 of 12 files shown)


src/main/resources/templates/entry.html
```diff
@@ -47,6 +47,12 @@ <h1 class="entry-header" th:text="${record.primaryKey}">Entry Name Header</h1>
     </div>
 </div>
 
+<div class="row">
+    <div class="small-12 large-6 columns">
+        <p>View all versions of this record <a th:href="${@uk.gov.register.presentation.HtmlViewSupport@extractLinkFromLinkHeader(#ctx.httpServletResponse.getHeader('Link'))}">here</a></p>
```
philandstuff, line 52: extracting the link from the link header feels fragile (because we might later add another link rel in another header) and wrong (because the view should be able to provide the link directly)

src/main/java/uk/gov/register/presentation/resource/HistoryResource.java
```diff
@@ -0,0 +1,61 @@
...
+            super(requestContext, "versions.html");
+            this.versions = versions;
+        }
+
+        public List<Version> getVersions() {
+            return versions;
+        }
+    }
+
+    private static class VersionJsonSerializer extends JsonSerializer<ListVersionView> {
```
philandstuff: This class name suggests it serializes `Version`, not `ListVersionView`...

src/main/java/uk/gov/register/presentation/resource/SearchResource.java
```diff
@@ -54,6 +54,7 @@ public SingleResultView findByPrimaryKey(@PathParam("primaryKey") String key, @P
         if (key.equals(registerPrimaryKey)) {
             Optional<DbRecord> record = queryDAO.findByKeyValue(key, value);
             if (record.isPresent()) {
+                setResponseHeader(key, value);
```
philandstuff: this method name is unclear

src/test/java/uk/gov/register/presentation/resource/HistoryResourceTest.java
```diff
@@ -0,0 +1,28 @@
...
+                return "localhost";
+            }
+        };
+
+        HistoryResource resource = new HistoryResource(requestContext, null);
+
+        try {
+            resource.history("someOtherKey", "value");
+            fail("Must fail");
+        } catch (NotFoundException e) {
```
philandstuff, line 24: you can use `@Test(expected = NotFoundException.class)` to avoid the try/catch block here

src/main/resources/templates/entry.html
om-sharma, line 52: After discussion we thought that it is ok to keep this way for now.

src/main/java/uk/gov/register/presentation/resource/HistoryResource.java
om-sharma: Renamed to ListVersionViewJsonSerializer in next commit.

src/main/java/uk/gov/register/presentation/resource/SearchResource.java
om-sharma: Renamed to setVersionHistoryLinkHeader in next commit.

src/test/java/uk/gov/register/presentation/resource/HistoryResourceTest.java
om-sharma, line 24: Should have try/catch block to confirm that the exception is thrown from the method call only.

PR #58, opened 2015-09-14
Add entry url

philandstuff:
this PR adds the `/entry/<serial-number>` url we've discussed IRL.  Nothing will link to it, the `/hash/<hash>` url hasn't been changed, etc; this PR just adds the new resource.

### DIFF
--- a/src/test/java/uk/gov/register/presentation/resource/SearchResourceTest.java
+++ b/src/test/java/uk/gov/register/presentation/resource/SearchResourceTest.java
@@ -101,7 +101,7 @@ public class SearchResourceTest {
 
     @Test
     public void findBySerial_findsEntryFromDb() throws Exception {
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/register/presentation/resource/SearchResourceTest.java
```diff
@@ -85,6 +100,51 @@ public void findByHash_throwsNotFoundWhenHashIsNotFound() {
...
+    @Test
+    public void findBySerial_reportsBadSerialNumberAsNotFound() throws Exception {
+        try {
+            resource.findBySerial("abcd");
+            fail("expected NotFoundException");
+        } catch (NotFoundException e) {
+            // success
+        }
+    }
+
```
om-sharma, line 146: should we have an unit test for all representations supported by findBySerial()?

PR #69, opened 2015-09-17
Extra content styles

robinwhittleton:
General template tidying and CSS updates

### DIFF
--- a/src/main/resources/templates/entries.html
+++ b/src/main/resources/templates/entries.html
@@ -9,12 +9,12 @@
     <h1 th:text="${entries.size()} + ' entries'"></h1>
     <table id="entries" th:if="${#lists.isEmpty(entries) == false}">
         <thead>
-        <tr>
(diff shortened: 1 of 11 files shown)


src/main/resources/templates/entries.html
```diff
@@ -9,12 +9,12 @@
...
     <table id="entries" th:if="${#lists.isEmpty(entries) == false}">
         <thead>
-        <tr>
-            <th>hash</th>
-            <th:block th:each="keyValue : ${entries.get(0).content}">
-                <th><a th:href="${@uk.gov.register.presentation.HtmlViewSupport@fieldLink(keyValue.key)}" th:text="${keyValue.key}"></a></th>
-            </th:block>
-        </tr>
+            <tr>
+                <th>Hash</th>
```
philandstuff, line 13: I think this is maybe better lower case as the fields are lower case too

PR #71, opened 2015-09-17
Add ADR to document why we removed kafka

philandstuff:
This commit adds an [Architecture Decision Record](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions) describing why we
chose to migrate away from kafka and zookeeper.  An Architecture
Decision Record is a document that records the context in which we made
a decision, the decision itself, and the forseen consequences of the
decision.  It is useful to record this information to ensure that if we
ever want to revisit a decision we can understand what the original
context was.


jabley:
I really like to see this. I find it very useful to write stuff down and understand what past me was thinking when I did something. It also encourages conscious decision-making, rather than unconscious acceptance.

dcarley:
Another :+1: for this process/format.

PR #73, opened 2015-09-18
Make register domain configurable

philandstuff:
This was hardcoded in a couple of places.  This commit allows us to use
a configurable value, and defaults to `beta.openregister.org` for the
moment.


om-sharma:
Closing it after discussion. We don't think it is need at the moment.

PR #74, opened 2015-09-18
Pagination

om-sharma:
This PR adds the next and previous page link for '/feed' page. 
 There are link header available in response which are used to render the next and previous page link. Changes are just rendering raw next and previous page links, design for these is still pending for which I have to talk to Stephen.


src/main/java/uk/gov/register/presentation/dao/RecentEntryIndexQueryDAO.java
```diff
@@ -37,6 +37,9 @@
             "GROUP BY (entry #>> ARRAY['entry',:key])) AS ii " +
             "WHERE i.id = ii.id " +
             "ORDER BY (entry #>> ARRAY['entry',:key]) DESC LIMIT :limit")
-    List<DbEntry> getLatestEntriesOfAllRecords(@Bind("key") String name,
-                                               @Bind("limit") int maxNumberToFetch);
+    List<DbEntry> getLatestEntriesOfRecords(@Bind("key") String name,
+                                            @Bind("limit") long maxNumberToFetch);
+
+    @SqlQuery("SELECT reltuples\\:\\:bigint FROM pg_class where relname='ordered_entry_index'")
+    long getTotalEntriesCount();
```
philandstuff: can this method name indicate that it's an estimated count?

src/main/java/uk/gov/register/presentation/resource/DataResource.java
```diff
@@ -50,19 +51,21 @@ public Response downloadTorrent() {
     @GET
     @Path("/feed")
     @Produces({MediaType.TEXT_HTML, MediaType.APPLICATION_JSON, ExtraMediaType.TEXT_YAML, ExtraMediaType.TEXT_CSV, ExtraMediaType.TEXT_TSV, ExtraMediaType.TEXT_TTL})
-    public EntryListView feed() {
-        return viewFactory.getEntryFeedView(
-                queryDAO.getAllEntries(ENTRY_LIMIT)
-        );
+    public EntryListView feed(@QueryParam("pageIndex") long pageIndex, @QueryParam("pageSize") long pageSize) {
```
philandstuff: what values will these `long`s be given if the query string isn't present? is it worth documenting this, or using a `Long` or `Optional<Long>` type to indicate that the values are optional?

src/main/java/uk/gov/register/presentation/resource/DataResource.java
```diff
@@ -50,19 +51,21 @@ public Response downloadTorrent() {
...
+        Pagination pagination = new Pagination(pageIndex, pageSize, queryDAO.getTotalEntriesCount());
+
+        List<DbEntry> entries = queryDAO.getAllEntries(pagination.pageSize(), pagination.offset());
```
philandstuff, line 61: you could pass the whole `Pagination` object to the DAO and use the [`@BindBean` annotation](http://jdbi.org/sql_object_api_argument_binding/) to pull out the size and offset values. Do you think that might be a little cleaner?

src/main/java/uk/gov/register/presentation/view/LinkHeaderExtractor.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    String extractLink(String linkRel) {
+        String linkHeader = httpServletResponse.getHeader("Link");
+
+        if (StringUtils.isEmpty(linkHeader)) {
+            return null;
+        }
+
+        return Stream.of(linkHeader.split(","))
```
philandstuff: ugh, I don't like that we're writing our own code to parse Link headers :(

src/main/resources/templates/feed.html
```diff
@@ -28,6 +28,9 @@ <h1 th:text="${entries.size()} + ' entries'"></h1>
         </tbody>
     </table>
 
+    <a th:if="${previousPage}" th:href="${previousPage}"> previousPage</a>
```
philandstuff: you have unnecessary leading spaces here

philandstuff:
done reviewing, can you address the comments above please?

src/main/java/uk/gov/register/presentation/resource/DataResource.java
om-sharma: Using 'Long' adds extra logic to handle the values. All the primitive types initialised with default values in jersey controller binding if the related parameters doesn't have any value. Are you suggesting to write javadoc?

src/main/java/uk/gov/register/presentation/resource/DataResource.java
om-sharma, line 61: I thought about that but there is some logic to construct the Pagination object. which kind of resets the pageSize and pageIndex values. To bind object as bean is comparatively complex.

src/main/resources/templates/feed.html
om-sharma: Removing in next commit.

src/main/java/uk/gov/register/presentation/dao/RecentEntryIndexQueryDAO.java
om-sharma: Renamed as suggested in next commit.

src/main/java/uk/gov/register/presentation/view/LinkHeaderExtractor.java
om-sharma: I also don't like it. Any suggestion?

src/main/java/uk/gov/register/presentation/resource/DataResource.java
philandstuff: just that I don't immediately know how missing values will be handled.  dropwizard already ensures that `Optional` query parameters can be processed appropriately, so there won't be any extra logic.

src/main/java/uk/gov/register/presentation/resource/DataResource.java
philandstuff, line 61: fair enough :+1:

philandstuff, line 61: wait, no, I don't understand. you need the logic anyway, but how does that have any bearing on `@BindBean`?

src/main/java/uk/gov/register/presentation/view/LinkHeaderExtractor.java
philandstuff: put the link url in the `View` object as well as in the response header, as I've suggested before

philandstuff:
I wonder if this could be refactored so we don't need setters on our View objects any more.  But this is good enough for now.

PR #77, opened 2015-09-21
Extracted next and previous page link in Pagination class.

om-sharma:
Extracted next and previous page link in Pagination class. Also rendering more details in pagination link

### DIFF
--- a/src/main/java/uk/gov/register/presentation/resource/Pagination.java
+++ b/src/main/java/uk/gov/register/presentation/resource/Pagination.java
@@ -4,13 +4,15 @@ package uk.gov.register.presentation.resource;
 import javax.ws.rs.BadRequestException;
 import java.util.Optional;
 
-class Pagination {
+public class Pagination {
 
     private final long pageIndex;
+    private final String resourcePath;
     private final long totalEntries;
     private final long pageSize;
 
-    Pagination(Optional<Long> pageIndex, Optional<Long> pageSize, long totalEntries) {
+    Pagination(String resourcePath, Optional<Long> pageIndex, Optional<Long> pageSize, long totalEntries) {
+        this.resourcePath = resourcePath;
         this.totalEntries = totalEntries;
 
         this.pageIndex = pageIndex.orElse(1L);
@@ -33,15 +35,28 @@ class Pagination {
         return pageIndex > 1;
     }
 
-    public long nextPageNumber() {
+    public long getNextPageNumber() {
         return pageIndex + 1;
     }
 
-    public long previousPageNumber() {
+    public long getPreviousPageNumber() {
         return pageIndex - 1;
+    }
+
+    public String getNextPageLink() {
+        return String.format("%s?pageIndex=%s&pageSize=%s", resourcePath, getNextPageNumber(), pageSize());
+    }
+
+    public String getPreviousPageLink() {
+        return String.format("%s?pageIndex=%s&pageSize=%s", resourcePath, getPreviousPageNumber(), pageSize());
     }
 
     public long pageSize() {
         return pageSize;
+    }
+
+    public long getTotalPages() {
+        return Math.round(Math.ceil(((double) totalEntries) / pageSize()));
+
     }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/register/presentation/resource/Pagination.java
```diff
@@ -4,13 +4,15 @@
...
-class Pagination {
+public class Pagination {
 
     private final long pageIndex;
+    private final String resourcePath;
     private final long totalEntries;
     private final long pageSize;
 
-    Pagination(Optional<Long> pageIndex, Optional<Long> pageSize, long totalEntries) {
+    Pagination(String resourcePath, Optional<Long> pageIndex, Optional<Long> pageSize, long totalEntries) {
```
philandstuff, line 14: again, the totalEntries variable should indicate that it's an estimate

src/main/java/uk/gov/register/presentation/resource/Pagination.java
```diff
@@ -33,15 +35,28 @@ public boolean hasPreviousPage() {
...
+
+    public String getPreviousPageLink() {
+        return String.format("%s?pageIndex=%s&pageSize=%s", resourcePath, getPreviousPageNumber(), pageSize());
+    }
+
     public long pageSize() {
         return pageSize;
     }
+
+    public long getTotalPages() {
```
philandstuff, line 58: and by extension this is an estimate of the total pages

src/main/resources/templates/feed.html
```diff
@@ -29,8 +29,16 @@ <h1 th:text="${entries.size()} + ' entries'"></h1>
         </tbody>
     </table>
 
-    <a th:if="${previousPageLink}" th:href="${previousPageLink}">previousPage</a>
-    <a th:if="${nextPageLink}" th:href="${nextPageLink}">nextPage</a>
+    <li th:if="${pagination.hasPreviousPage()}" class="previous">
```
philandstuff: `<li>` isn't valid here, because you're not in a list.  What are you trying to do?

src/main/java/uk/gov/register/presentation/resource/Pagination.java
om-sharma, line 14: For pagination class it is number of totalEntries only. you may chose to pass estimated entries or total entries from the caller. total entries in pagination does not mean the number of records in database, it means total entries to paginate on.

src/main/java/uk/gov/register/presentation/resource/Pagination.java
om-sharma, line 58: same as above comment.

src/main/resources/templates/feed.html
om-sharma: You said that you will ask Robin to look at this so I didn't give much time to look at the html. I am not sure that we need `<li>`. I copied the pieces from alpha so forgot to copy `<ul>` part. Although I agree these `<li>` must be covered in `<ul>`.

src/main/java/uk/gov/register/presentation/resource/Pagination.java
philandstuff, line 14: does this mean that the pagination may fail to render all pages if the estimate is wrong?

src/main/resources/templates/feed.html
philandstuff: can you at least make it valid before I merge it? ie add the `<ul>`s in?

src/main/java/uk/gov/register/presentation/resource/Pagination.java
om-sharma, line 14: Pagination class is used for a request, so the number of entries is snapshot at the time of this request. number of entries are calculated for every request. 
Estimated entries are never greater than the total entries in table, so given the way it is used, it should never fail.

src/main/resources/templates/feed.html
om-sharma: Done!

src/main/java/uk/gov/register/presentation/resource/Pagination.java
philandstuff, line 14: discussed IRL. we have a strategy to use fast, exact queries so we don't need to estimate after all

PR #78, opened 2015-09-21
Add register configuration

philandstuff:
This PR adds RegistersConfiguration -- similar to FieldsConfiguration but for registers.

It also ensures that we use RegistersConfiguration to work out which fields to render, rather than guessing based on the first entry in a list.  This means that, for example, we can properly support optional fields now, because we know what fields to expect when rendering tables (csv, tsv, html).


om-sharma:
Can you please add some sanity tests for the html pages changed in the PR? 

I received 500 response when browsed the entry page.

PR #84, opened 2015-09-28
Slow current query

om-sharma:
Using new table 'register_entries_count' to get number of all entries in register. this use replace the estimated count query.
Using current_keys table to fetch the current entries to make '\current' call faster.


src/test/resources/fixtures/list.csv
```diff
@@ -1,3 +1,3 @@
 entry,address,area,country,latitude,locality,longitude,postcode,property,street,town
-2,67890,value2,,,,,,,The Entry 2,
-1,12345,value1,,,,,,,The Entry 1,
+1,12345,value1,,,,,,,The Entry 1,town1
```
philandstuff: why has this order changed? don't we want most-recent-first?

PR #90, opened 2015-10-12
Reload local copy of registers list

om-sharma:
registers data was changed recently, synced the local copy of beta registers.

Prod Field register (http://field.prod.openregister.org/current) was showing old fields which were [field, fields, datatype, register,text].

### DIFF
--- a/src/main/resources/config/registers.yaml
+++ b/src/main/resources/config/registers.yaml
@@ -1,10 +1,8 @@
 ---
-- hash: "c17e5f868dac5872a89707ac147a07f24f3988b7"
+- hash: "1156f7416b4467e445a3d0427cd839be4eea3d23"
   entry:
-    copyright: "Contains Ordnance Survey data © Crown copyright & database right\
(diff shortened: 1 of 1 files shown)


philandstuff:
can you explain what you mean by "synced the local copy of beta registers"? did you get this data from the prod register register or the alpha register register? It's not clear from the commit message.

Also I don't know what you mean by "Prod Field register was showing old fields".  What should it have been showing?

om-sharma:
List of beta registers are loaded from the file "registers.yaml" which is in the presentation source code. The content of this file was outdated. It was needed to be synced with the registers data which is currently in https://github.com/openregister/registers/tree/master/data/Register. This data was recently  reviewed and edited by psd and Adam, I think.

I have already created field register in prod environment and loaded the data into it. If you browse http://field.prod.openregister.org/current, you will see header for entries. this header shows fields in field register which are loaded from "registers.yaml" file. this is the old value of fields field of field register.

philandstuff:
aha, I think I understand now

PR #91, opened 2015-10-13
Render fields with cardinality "MANY"

philandstuff:
The register register has a "fields" field which is a list of fields. This PR adds the ability to render such fields with a cardinality of MANY, across all representations.

Things of note:
- FieldValue.value() has been renamed FieldValue.getValue().  This is
  largely for EntryConverterTest, which can use the
  samePropertyValuesAs() matcher to compare LinkValue() types
- we use semicolons to separate values in CSV and TSV representations;
  this is slightly iffy. The idea was copied over from the alpha.

Things not done:
- I wanted to make FieldValue.isLink() and FieldValue.isList() into
  default methods, but there's a bug in OGNL (used by thymeleaf) which
  hides default methods from the templates.  This is fixed in OGNL
  3.0.12 (we're currently on 3.0.8).

See individual commits for more context.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/Cardinality.java
+++ b/src/main/java/uk/gov/register/presentation/Cardinality.java
@@ -2,9 +2,8 @@ package uk.gov.register.presentation;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
-import com.google.common.collect.ImmutableMap;
 
-import java.util.Map;
+import java.util.stream.Stream;
 
 public enum Cardinality {
     ONE("1"),MANY("n");
@@ -15,11 +14,6 @@ public enum Cardinality {
         this.id = id;
     }
 
-    private static final Map<String,Cardinality> cardinalityMap = ImmutableMap.of(
-            ONE.getId(),  ONE,
-            MANY.getId(), MANY
-    );
-
     /**
      * this method shouldn't be needed -- the @JsonValue annotation has a special case for enums that
      * means it can be used for both serialization and deserialization -- but there's a bug.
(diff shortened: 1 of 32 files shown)


src/main/java/uk/gov/register/presentation/Cardinality.java
```diff
@@ -0,0 +1,37 @@
...
+    );
+
+    /**
+     * this method shouldn't be needed -- the @JsonValue annotation has a special case for enums that
+     * means it can be used for both serialization and deserialization -- but there's a bug.
+     * See <a href="https://github.com/dropwizard/dropwizard/issues/699">dropwizard issue #699</a>
+     */
+    @JsonCreator
+    public static Cardinality fromId(String id) {
+        return cardinalityMap.get(id);
```
om-sharma: we can change the implementation as `Stream.of(Cardinality.values()).filter(c -> c.getId().equals(id)).findFirst().get();`,  which will remove the cardinalityMap presence.

om-sharma:
Can you please rebase your branch with master. I can not locally see that the fields are rendered properly as a list. This is because of the 'registers.yaml' file is outdated here which is now changed in master.

src/main/java/uk/gov/register/presentation/Cardinality.java
philandstuff: I'm ambivalent about this. Happy to change it if you prefer it but I'm not bothered.

philandstuff:
rebased.

src/main/java/uk/gov/register/presentation/Cardinality.java
om-sharma: I am more concerned about duplicating all enum objects in separate map, which is already provided by enums in some form.

PR #93, opened 2015-10-14
Using prod register for linking everywhere

om-sharma:
Till now were linking to alpha registers. Changes in this PR allows us to link to the beta registers.


philandstuff:
needs rebase

om-sharma:
We will migrate all prod.openregister.org to  openregister.org very soon so doing this is not worth.

PR #94, opened 2015-10-15
Quick primary key search 

om-sharma:
Search by primary key was slow, so to make it faster now using sub query from current_keys table.


philandstuff:
:+1: :+1: :whale:

PR #95, opened 2015-10-15
bug: html view for register register was broken

om-sharma:
After introducing the ListValue class html view of register register was broken because it is only register which has fields whose value is a list. getValue() method of ListValue was not implemented so html was unable to render.
   Changes in this commit renders fields as plain text list of fields, these are not linked to field register. There is more work needed to achieve this.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/ListValue.java
+++ b/src/main/java/uk/gov/register/presentation/ListValue.java
@@ -2,6 +2,7 @@ package uk.gov.register.presentation;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/register/presentation/ListValue.java
```diff
@@ -21,7 +22,7 @@ public boolean isList() {
     @Override
     @JsonIgnore
     public String getValue() {
-        throw new UnsupportedOperationException();
+        return "[ " + String.join(", ", Lists.transform(elements, FieldValue::getValue)) + " ]";
```
philandstuff, line 25: I don't think it's this object's job to render html - and in particular, this won't render the LinkValues as links

om-sharma, line 25: I agree, this object should not render the html. I see getValue() method kind of toString() method for this object. 
It does not renders linkValues as links and we need more work to do that (stated in commit message). This is quick solution to resolve 500 response from register register.

om-sharma:
discussed IRL: closing it for now, will do list of fields rendering in html

PR #96, opened 2015-10-19
Add default cache-control header of no-transform

philandstuff:
This adds a `Cache-control: no-transform` header.  When browsing
registers from certain mobile networks, there are middleboxes which try
to "optimize" traffic by replacing `<link>` tags with inline `<style>`s;
however this causes a Content-Security-Policy violation.  The
`no-transform` directive politely asks them not to do that.  Rumours on
the internet are that tools like Bytemobile will respect this.


psd:
Works for me! (giffgaff / O2) :+1:  

.. ah, but this hasn't been merged, so it's an intermittent issue. Great.

PR #97, opened 2015-10-20
Render markdown

philandstuff:
See commits for context

### DIFF
--- a/src/main/resources/templates/home.html
+++ b/src/main/resources/templates/home.html
@@ -8,7 +8,7 @@
     <div th:replace="main.html::phase"></div>
     <div class="grid-row">
         <div class="column-two-thirds">
-            <h1 class="homepage-intro" th:text="${register.text}">The register of UK government registers.</h1>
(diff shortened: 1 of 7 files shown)


src/main/resources/templates/home.html
```diff
@@ -8,7 +8,7 @@
     <div th:replace="main.html::phase"></div>
     <div class="grid-row">
         <div class="column-two-thirds">
-            <h1 class="homepage-intro" th:text="${register.text}">The register of UK government registers.</h1>
+            <div class="homepage-intro" th:utext="${@uk.gov.register.presentation.HtmlViewSupport@renderMarkdown(register.text)}">The register of UK government registers.</div>
```
om-sharma, line 12: `th:utext` did you mean `th:text`

philandstuff, line 12: no. `th:utext` means "unescaped text" which is what we want if we want the tags created by the markdown render to be rendered as tags rather than escaped as entities like `&lt;`.

om-sharma, line 12: Ok.  didn't know about it. :+1:

PR #98, opened 2015-10-20
Register register html representation bug

om-sharma:
getValue() method in ListValue class was not implemented which fails to render registers where field value is a list value. Rendering list value types in html.

### DIFF
--- a/src/main/sass/main.scss
+++ b/src/main/sass/main.scss
@@ -157,6 +157,7 @@ dd a {
 
 .field-list ul {
   padding-left: 0;
+  margin-top: 0;
 }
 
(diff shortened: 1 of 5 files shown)


src/main/sass/main.scss
```diff
@@ -176,6 +176,25 @@ dd a {
...
+}
+
+.field-list-in-table li {
+  font-family: monospace;
+  font-size: 14px;
+  display: inline-block;
+ }
+
+.field-list-in-table li:after {
+  content: ", ";
```
philandstuff: is there a way you could extract this stuff to a common class?

PR #102, opened 2015-10-22
Linking register's provided by text to public body register

om-sharma:
Linking register's provided by text to public body register

### DIFF
--- a/src/main/java/uk/gov/register/presentation/HtmlViewSupport.java
+++ b/src/main/java/uk/gov/register/presentation/HtmlViewSupport.java
@@ -25,11 +25,11 @@ public class HtmlViewSupport {
     }
 
     public static String fieldLink(String fieldName) {
-        return UriBuilder.fromPath("http://field.openregister.org/field").path(fieldName).build().toString();
+        return new LinkValue("field", fieldName).link();
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/register/presentation/HtmlViewSupport.java
```diff
@@ -28,6 +28,10 @@ public static String fieldLink(String fieldName) {
         return UriBuilder.fromPath("http://field.openregister.org/field").path(fieldName).build().toString();
     }
 
+    public static String publicBodyLink(String publicBodyId){
+        return UriBuilder.fromPath("http://public-body.openregister.org/public-body").path(publicBodyId).build().toString();
```
philandstuff: does this url have to be hardcoded? can't we use the link rendering code we use elsewhere for field links?

om-sharma: Are you talking about LinkValue? We can not use LinkValue for this purpose as that is a FieldValue. we can use common template for register urls in the application so that changes are at one place but somewhere we have to hardcode public-body register.

om-sharma:
Updated to use LinkValue instead of duplicating the logic

PR #104, opened 2015-10-23
Resolved Bug: /feed returns 404 when no records in database.

om-sharma:
There was a check in Pagination class constructor which verifies that entries exist for given page number and size. If there is no entry for the given request parameters We throw NotFoundException. This check was causing 404 on /feed when you browse first page and there is no entries in database. In this case we should not throw 404 but render the page with zero entries.
The change in constructor confirms that the check is made only when the request is for a page number 2 or more. Now skipping the check for first page request.
There was a functional test failing because of comparision failure of utf-8 v/s UTF-8. So capitalized utf-8 to UTF-8 everywhere to resolve confusion in tests.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/representations/ExtraMediaType.java
+++ b/src/main/java/uk/gov/register/presentation/representations/ExtraMediaType.java
@@ -3,14 +3,14 @@ package uk.gov.register.presentation.representations;
 import javax.ws.rs.core.MediaType;
 
 public class ExtraMediaType {
-    public static final String TEXT_HTML = "text/html; charset=utf-8";
+    public static final String TEXT_HTML = "text/html; charset=UTF-8";
 
-    public static final String TEXT_CSV = "text/csv; charset=utf-8";
-    public static final MediaType TEXT_CSV_TYPE = new MediaType("text", "csv", "utf-8");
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/register/presentation/representations/ExtraMediaType.java
```diff
@@ -3,14 +3,14 @@
 import javax.ws.rs.core.MediaType;
 
 public class ExtraMediaType {
-    public static final String TEXT_HTML = "text/html; charset=utf-8";
```
philandstuff, line 6: what's the purpose of this change?

philandstuff:
what's the fix? is it the utf-8 -> UTF-8 thing or the Pagination thing? neither of them look like they should make any significant change :/

philandstuff:
I think we should either standardize on lowercase everywhere, or make the functional test perform case-insensitive comparison.  While [no distinction is made between uppercase and lowercase letters in charset names](http://www.iana.org/assignments/character-sets/character-sets.xhtml), in practice lowercase is far more common.

philandstuff:
after discussing IRL it looks like it's not worth messing about with the utf-8 stuff. it works, leave it alone.

PR #107, opened 2015-11-03
All database queries must be written in lowercase characters

om-sharma:
All queries must be written in lowercase characters to remove any confusion and probable bug.

 Postgres entities are created as all lowercase characters. we faced a problem in indexer app in which where clause was specifying an index name in uppercase and the condition was never met (https://github.com/openregister/indexer/pull/23).


om-sharma:
Closing the pull request as there is no worth doing it here.

PR #108, opened 2015-11-04
Render total records on home page

om-sharma:
NOTE: Do not merge this PR before successful merge of PR https://github.com/openregister/indexer/pull/25 in indexer repo. This is required because the total_records table is created by indexer app.

This PR renders total records on register home page, Using total_records table to fetch the total records in register.


src/main/java/uk/gov/register/presentation/view/ViewFactory.java
```diff
@@ -55,7 +59,7 @@ public ThymeleafView thymeleafView(String templateName) {
     }
 
     public HomePageView homePageView() {
```
philandstuff: shouldn't we pass in the data to this method, rather than having the ViewFactory pass the DAO into the View?

PR #115, opened 2015-11-13
Rename all the urls

philandstuff:
Rename:
- /feed to /entries
- /current to /records
- ?pageIndex to ?page-index
- ?pageSize to ?page-size

Also remove /all and /latest, which were old redirects that we don't need anymore.


philandstuff:
Tested the widgets-demo against this merged PR on the preview countries register.  It happily follows the 301 redirect and fetches the countries.  This means we can deploy to production without coordinating any sort of change to the demos.

PR #119, opened 2015-11-23
Add digital proof to record page

philandstuff:
Similar to #118.  As the proof is basically the same, I extracted a thymeleaf fragment to capture the digital proof part (and pulled the hardcoded constants into EntryView because defining them in Thymeleaf was too horrible).


philandstuff:
argh fails build. I'll deal with it.

PR #120, opened 2015-11-23
Renamed feed.html to entries.html and current.html to records.html

om-sharma:
Also renamed view methods to improve readability


philandstuff:
:+1:

PR #122, opened 2015-11-24
page-size more than 5000 is not accepted now

om-sharma:
- earlier you could type an url which has very large page size which causes the db queries to run for long time. Now limited page size to 5000 so that app can fail fast and database is not loaded with heavy queries.

### DIFF
--- a/src/main/resources/templates/400.html
+++ b/src/main/resources/templates/400.html
@@ -8,7 +8,7 @@
 <main id="main" role="main">
     <div th:replace="main.html::phase"></div>
     <h1>Bad Request</h1>
-    <p>Please check the request parameters.</p>
+    <p th:text="${exception.message}"></p>
     <p th:replace="copyright.html::copyright"></p>
 </main>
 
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/register/presentation/resource/PaginationTest.java
```diff
@@ -12,6 +12,11 @@
 public class PaginationTest {
 
     @Test(expected = BadRequestException.class)
+    public void construct_throwsExceptionForPageSizeGreaterThan10000() {
```
philandstuff: test name lies :)

src/main/resources/templates/400.html
```diff
@@ -0,0 +1,17 @@
...
+
+<html xmlns:th="http://www.thymeleaf.org" xmlns="http://www.w3.org/1999/xhtml">
+<head th:include="main.html::head"></head>
+<body>
+<header th:replace="main.html::header"></header>
+
+<main id="main" role="main">
+    <div th:replace="main.html::phase"></div>
+    <h1>Bad Request</h1>
+    <p>Please check the request parameters.</p>
```
philandstuff: Should we maybe be clearer about the exact error rather than having a generic response here?

PR #129, opened 2015-11-30
Frontend fixups

philandstuff:
After feedback from @robinwhittleton. See individual commits for more detail.


robinwhittleton:
👍

PR #130, opened 2015-11-30
Create field object by taking all values as string.

om-sharma:
it simplifies the Cardinality enum code.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/config/Field.java
+++ b/src/main/java/uk/gov/register/presentation/config/Field.java
@@ -5,20 +5,22 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.commons.lang3.StringUtils;
 import uk.gov.register.presentation.Cardinality;
 
+import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 public class Field {
     final String fieldName;
     final String datatype;
     final Optional<String> register;
-    final Cardinality cardinality;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/register/presentation/config/Field.java
```diff
@@ -32,7 +34,7 @@ public Field(@JsonProperty("field") String fieldName,
     }
 
     public Cardinality getCardinality() {
-        return cardinality;
+        return Stream.of(Cardinality.values()).filter(c -> Objects.equals(c.getId(), cardinality)).findFirst().get();
```
philandstuff, line 37: wouldn't it be better to do this in the constructor rather than constructing a potentially-invalid object and blowing up only when someone calls getCardinality?

philandstuff, line 37: also doesn't dropwizard/dropwizard#699 fix the problem you're trying to solve?

om-sharma, line 37: Fail fast is good but in this particular case I am not very keen to do it because we are still in development mode and there are cases where we add some data entry which can not be deserialized because of which the register is down. 

I guess first we should make sure that the data is clean and validated.

Also dropwizard/dropwizard#699 is still an open issue, the changes above are good to simplify the cardinality enum code.

om-sharma:
Closing this PR  as this is not a important thing to do at this moment.

PR #131, opened 2015-12-01
Fix column widths on single entry pages

philandstuff:
This fixes our column width to one-third / two-thirds for field name and
value respectively.  This applies to the record and entry pages.

### DIFF
--- a/src/main/resources/templates/fragments/entry-table.html
+++ b/src/main/resources/templates/fragments/entry-table.html
@@ -4,6 +4,10 @@
 <body>
 <div th:fragment="entry-table(content)">
     <table class="entry">
+        <colgroup>
+            <col class="field-column" />
+            <col class="value-column" />
+        </colgroup>
         <thead>
             <tr>
                 <th>Field</th>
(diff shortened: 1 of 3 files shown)


philandstuff:
@robinwhittleton would you mind taking this?

src/main/resources/templates/fragments/entry-table.html
```diff
@@ -4,6 +4,10 @@
 <body>
 <div th:fragment="entry-table(content)">
     <table class="entry">
+        <colgroup>
+            <col class="field-column" />
```
philandstuff, line 8: fyi our templating language (thymeleaf) is xml based so these trailing `/` characters are required. they won't appear in the html5 output.

philandstuff, line 8: sorry, this is nonsense. the `/` will appear in the output but doesn't seem to cause problems.

robinwhittleton, line 8: [No problems with self-closing tags in HTML5](http://www.w3.org/TR/html51/syntax.html#start-tags)

robinwhittleton:
LGTM

src/main/resources/templates/fragments/entry-table.html
philandstuff, line 8: great, good to know!

PR #133, opened 2015-12-04
Pagination header

om-sharma:
Restyled page navigation links on records and entries page.

screenshot 2015-12-04 12 05 56
screenshot 2015-12-04 12 06 40
screenshot 2015-12-04 12 07 31


src/main/java/uk/gov/register/presentation/resource/Pagination.java
```diff
@@ -43,6 +43,23 @@ long offset() {
         return (pageIndex - 1) * pageSize;
     }
 
+    public long getTotalEntries() {
+        return totalEntries;
+    }
+
+    public boolean isFirstPageAndShowsAllEntries() {
```
philandstuff: could this have a better name? `isOnlyPage` or `isSolePage` or something like that?

src/main/sass/_page-navigation.scss
```diff
@@ -0,0 +1,105 @@
+.previous-next-navigation {
```
philandstuff: Can you document where this came from?

philandstuff:
:+1: other than the minor comments above

src/main/sass/_page-navigation.scss
philandstuff: maybe steal from https://github.com/alphagov/static/blob/master/app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss ?

philandstuff: spoke to @dsingleton who said that pulling things in from govuk-component might not be the best idea as it's very specific to GOV.UK's own needs.

If there were a navigation widget in govuk-element it might have been nice to use it, but there isn't.

So, this is fine for now.

ping @robinwhittleton who might be interested in this discussion

src/main/resources/templates/records.html
```diff
@@ -7,9 +7,7 @@
 <main id="main" role="main">
     <div th:replace="main.html::phase"></div>
     <div class="grid-row">
-        <div class="column-two-thirds">
-            <h1 th:text="${entries.size()} + ' records'"></h1>
```
robinwhittleton, line 11: We’re losing our `h1` for the page here. Is there something replacing it?

src/main/sass/_page-navigation.scss
dsingleton: Hi, had a look at this PR in a bit more detail after our discussion.

I think re-using the naming convention (`<div class="govuk-previous-and-next-navigation">`) would be a bad idea, as it might suggest a direct link between the code-bases, where one doesn't exists, introducing confusion. Changing the root scope/class name, and keeping the rest of the markup/CSS, would be :+1:, possible with a "based on <git link> comment".

I'd recommend using the component markup/CSS over other GOV.UK implementations as it's likely to be more modern/polished (for example, the component has better accessibility support). That would also be the best starting point for adding to elements.

src/main/sass/_page-navigation.scss
```diff
@@ -0,0 +1,105 @@
...
+    width: 100%;
+    padding-left: 0;
+}
+
+@media only screen and  (max-width: 64em) {
+    .previous-next-navigation {
+        width: 90%;
+    }
+
+    p.download {
```
philandstuff: this isn't needed.

src/main/sass/_page-navigation.scss
```diff
@@ -0,0 +1,105 @@
+.previous-next-navigation {
+    position: relative;
+    overflow: hidden;
+    margin-left: 0px;
+    width: 100%;
+    padding-left: 0;
+}
+
+@media only screen and  (max-width: 64em) {
+    .previous-next-navigation {
```
philandstuff: can you add a screenshot demonstrating this bit?

src/main/sass/_page-navigation.scss
```diff
@@ -0,0 +1,105 @@
...
+		font-family: "nta", Arial, sans-serif;
+		font-size: 23px;
+		line-height: 1.11111;
+		font-weight: 400;
+		text-transform: none;
+		display: inline-block;
+		position: static;
+	}
+}
+@media (min-width: 641px) and (max-width: 640px) {
```
philandstuff: what does this media query mean?

src/main/sass/_page-navigation.scss
```diff
@@ -0,0 +1,105 @@
+.previous-next-navigation {
+    position: relative;
+    overflow: hidden;
+    margin-left: 0px;
+    width: 100%;
+    padding-left: 0;
+}
+
+@media only screen and  (max-width: 64em) {
```
philandstuff: why is this `64em` rather than `640px` elsewhere in this file?

src/main/sass/_page-navigation.scss
```diff
@@ -0,0 +1,105 @@
...
+    margin: 0;
+    width: 50%;
+    float: left;
+}
+@media (max-width: 640px) {
+    .previous-next-navigation li {
+    font-size: 18px;
+    line-height: 1.11111;
+}
+}.previous-next-navigation li.next {
```
philandstuff: the formatting's a bit poor here

src/main/sass/_page-navigation.scss
```diff
@@ -0,0 +1,105 @@
...
+    float: right;
+    text-align: right;
+}
+.previous-next-navigation li.next a {
+    padding: 15px 35px 15px 0;
+}
+.previous-next-navigation li.next a:before {
```
philandstuff: it looks like there's a lot of opportunity to take advantage of sass's nesting features here.

src/main/resources/templates/records.html
om-sharma, line 11: Yes `h1` is replaced by text i.e. `1-100 of 1000 records` or `56 records`. Have a look at the attached png's with the PR.

PR #139, opened 2015-12-08
Setting dev environment with 1000 entries from school datatbase

om-sharma:
We have `run-application.sh` which creates the db schema and then starts application in dev environment. This script was not update with the schema changes also there was no data available for to browse application in dev environment.

Now changes in PR will create school database on dev machine and will load 1000 entries from school register.


om-sharma:
Closing this PR as we need more work.

PR #141, opened 2015-12-09
Default the business field to use 'company' as the default register for linking

AndrewGorton:


philandstuff:
specifically, this updates fields.yaml to the latest version from field.openregister.org

PR #145, opened 2015-12-10
Logos for organisations

philandstuff:
This uses code from alphagov/static to render logos alongside
organisation names on the front page and the top-right of the list and
single entry pages.

The ViewFactory is responsible for calling the GOV.UK organisations API
to fetch the logo, colour and wording for each organisation.

It seems that Thymeleaf strips newlines from text substituted on to the
page, so to get the newlines in the organisation branding to render
properly, I had to manually replace them with `<br/>` tags.


src/main/java/uk/gov/register/presentation/config/PresentationConfiguration.java
```diff
@@ -10,6 +12,10 @@
 public class PresentationConfiguration extends Configuration {
     @Valid
     @NotNull
+    private JerseyClientConfiguration jerseyClient = new JerseyClientConfiguration();
```
om-sharma, line 17: Do you want to provide ClientConfiguration parameters optionally from configuration? if yes @JsonProperty is missing and @NotNull annotation is not required. If not then any specific reason this object created in PresentationConfiguration?

src/main/java/uk/gov/register/presentation/view/ViewFactory.java
```diff
@@ -64,14 +71,22 @@ public BadRequestExceptionView badRequestExceptionView(BadRequestException e) {
...
+        return new ListVersionView(requestContext, getCustodian(), getBranding(), versions);
     }
 
     private PublicBody getCustodian() {
         return publicBodiesConfiguration.getPublicBody(requestContext.getRegister().getRegistry());
     }
+
+    private GovukOrganisation.Details getBranding() {
+        String custodianId = requestContext.getRegister().getRegistry();
+        return client.target("https://www.gov.uk/api/organisations/" + custodianId)
```
om-sharma: handle failure gracefully!

philandstuff:
@om-sharma updated to handle failure gracefully.

PR #146, opened 2016-01-04
bugfix: preserve data format in pagination links

philandstuff:
There was a bug where fetching a specific data format such as
`/records.json` would provide pagination links that stripped the data
format from the path -- eg `/records?page-index=2`.

This commit fixes that bug by using query-string-relative URLs instead:
ie a relative URL that begins with a question mark `?` preserves the
existing URL up to the path and only replaces the query string. See
https://url.spec.whatwg.org/#relative-state for more on this.

This means we can get rid of some code: the Pagination no longer needs
to know the resource path of the paginated resource.


om-sharma:
Good Job!!  :thumbsup:

PR #148, opened 2016-01-12
UrlEncode the hash for hrefs

AndrewGorton:
"View All Entries" uses the entry hash as the target for the link, but the Base64 dictionary includes character instructions ("/", "+" and "="), thus breaking the link target.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/Version.java
+++ b/src/main/java/uk/gov/register/presentation/Version.java
@@ -1,8 +1,16 @@
 package uk.gov.register.presentation;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/register/presentation/Version.java
```diff
@@ -12,4 +20,13 @@ public Version(int serialNumber, String hash) {
         this.hash = hash;
         this.serialNumber = serialNumber;
     }
+
+    public String getUrlEncodedHash() {
```
philandstuff, line 24: can we have a test for this?

AndrewGorton:
Overtaken by recent discussions.

PR #149, opened 2016-01-14
Load test

om-sharma:
This PR provides has all the required scripts and library to run the load test. It uses gatling tool to run the tests. There is only one test scenario which I used and is committed in `user-files/simulations/countryregister` folder. 

note: There is still more work required to run different sample from the command line which can be done when we run load tests again.


om-sharma:
created a separate repository for this

PR #150, opened 2016-01-15
Removed /current and /feed resources

om-sharma:
These were not used anywhere and simply redirecting to /records and /entries .

### DIFF
--- a/src/main/java/uk/gov/register/presentation/resource/DataResource.java
+++ b/src/main/java/uk/gov/register/presentation/resource/DataResource.java
@@ -75,20 +75,6 @@ public class DataResource {
         return viewFactory.getRecordsView(queryDAO.getLatestEntriesOfRecords(pagination.pageSize(), pagination.offset()), pagination);
     }
 
-    @GET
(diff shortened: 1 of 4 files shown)


philandstuff:
@om-sharma I found a usage of `/feed`: https://github.com/openregister/organisations-demo/blob/master/manage.py#L22

philandstuff:
(found using a [github search](https://github.com/search?utf8=%E2%9C%93&q=user%3Aopenregister+%22feed%22&type=Code&ref=searchresults))

philandstuff:
@psd can you confirm if it's okay to remove `/current` and `/feed`? these are old URLs which redirect to `/records` and `/entries` respectively. We used the redirect to ease the transition to the new names. We'll make sure that none of the demos still use these URLs.

om-sharma:
@psd: Organisations demo is not in use. Is it ok to delete the heroku app?

om-sharma:
@philandstuff: removed the use of `/feed` from organisations demo repository.

PR #153, opened 2016-01-22
406 response when requested format is not supported

om-sharma:
Returning 406 response now instead of accidental 500 when the resource can not return response of requested type in accept headers.

http://register.openregister.org/register/food-premises/history.yaml resource currently returns 500 response because we don't support yaml format for this resource. jetty throws exception in this case which is handled by ThrowableExceptionMapper which converts it in InternalServerError. 

Ideally we should be able to convert resources in all the formats. this requires more work, but we must not convert any WebApplicationException in InternalServerError.


src/main/java/uk/gov/register/presentation/resource/ClientErrorExceptionMapper.java
```diff
@@ -0,0 +1,45 @@
...
+                .build();
+    }
+
+    private ThymeleafView getEntityFor(ClientErrorException exception) {
+        if (exception instanceof BadRequestException) {
+            return viewFactory.badRequestExceptionView((BadRequestException) exception);
+        }
+
+        if (exception instanceof NotFoundException || exception instanceof NotAcceptableException) {
+            return viewFactory.thymeleafView("404.html");
```
AndrewGorton, line 40: Shouldn't we report a [406](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.7) for Not Acceptable?

om-sharma, line 40: The 406 response is returned but html page displays "the page is not found" so reusing the `404.html` template. In case we want some other display text we can create new template. 
If it is confusing we can rename the template file.

philandstuff, line 40: tl;dr: I think we should merge this as it is. It's not perfect, but it's good enough for the current cost/benefit tradeoff.  There are two warts as a result:
- we raise a 406 in places we should really raise a 404 (say, at /download.csv)
- whenever we raise a 406 error, the response body is the same as a 404.

but fixing these would be a lot of effort for relatively little gain.

Long version:
We support getting different representations via both http content negotiation and different URLs for different representations. That is, you can fetch a json record either via GETting /record/school/1234.json or by GETting /record/school/1234 with an Accept: header of `application/json`.

There is an interesting question of what we should return when a given representation is not applicable for a particular resource -- for example, something like /download.csv. The /download resource is mostly a textual page which points you to the various ways to download all of the register, but the /download resource itself does not support formats like csv.

In my view, what should happen is this:
- if a user GETs /download.csv, they should see a 404, because they asked for a resource which doesn't exist
- if a user GETs /download with an Accept: header of `text/csv`, they should see a 406, because they asked for a resource which does exist but which doesn't have the representation asked for.

We use UriConnegFilter to manage the dispatch from URLs with representation suffixes (eg `.yaml`,`.json`) to resource URLs. The way the filter works is to translate an incoming request for `/foo.json` into a request for `/foo` with an Accept: header of `application/json`.  We then let jersey's content negotiation mechanisms kick in to play in order to convert whatever View object we return from our resource into an appropriate response serialization format.

The problem is that UriConnegFilter is not very smart when a given representation is not acceptable to a given resource. For example, visiting /download.csv will be translated into a request for /download with an Accept header of `text/csv; charset=utf-8`. The download resource will return a 406 Not Acceptable for this requested content type, which UriConnegFilter passes straight back to the user agent.

To implement this correctly, we would have to have a version of UriConnegFilter which was able to detect a 406 response from the upstream resource class, and convert it to a 404 response to the downstream user agent.  That way, /download.csv would return a 404 correctly, and /download with Accept: text/csv would return 406.

As things are, we return a 406 in all of these cases. Because we're sometimes returning a 406 when a 404 would be more correct, we've also made the 406 response body say "not found" in the same way that the 404 response body does.

Since the fix would involve either fixing UriConnegFilter (and feeding the change back upstream to jersey) or forking UriConnegFilter, that seems a lot of work for what is some fairly pedantic adherence to proper http, and in error cases only for that matter.

PR #154, opened 2016-01-25
implemented '/proof/certificate-transparency' resource

om-sharma:
Implemented resource which returns the signed tree head of the register to prove the integrity of data in register.

The resource returns the following json response:

```
{
  "tree_size": "9803348",
  "timestamp": "1447421303202",
  "sha256_root_hash": "JATHxRF5gczvNPP1S1WuhD8jSx2bl+WoTt8bIE3YKvU=",
  "tree_head_signature":
  "BAMARzBFAiEAkKM3aRUBKhShdCyrGLdd8lYBV52FLrwqjHa5/YuzK7ECIFTlRmNuKLqbVQv0QS8nq0pAUwgbilKOR5piBAIC8LpS"
}
```

The register proof view does not have any html template yet so it always returns json response. We need more clarity over the html pages for the resources. psd and @stephenjoe1 are working on it.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/SignedTreeHead.java
+++ b/src/main/java/uk/gov/register/presentation/SignedTreeHead.java
@@ -1,0 +1,18 @@
+package uk.gov.register.presentation;
+
+public class SignedTreeHead {
+    public final int tree_size;
+
+    public final long timestamp;
+
+    public final String sha256_root_hash;
+
+    public final String tree_head_signature;
+
+    public SignedTreeHead(int tree_size, long timestamp, String sha256_root_hash, String tree_head_signature) {
+        this.tree_size = tree_size;
+        this.timestamp = timestamp;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/register/presentation/SignedTreeHead.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.register.presentation;
```
philandstuff: I wonder if we might want to take these classes in to a separate package for CT stuff? maybe u.g.r.proofs.certificate-transparency?

philandstuff:
This is safe because openregister/indexer#34 ensures that the `sth` table always exists, for the postgres or CT backends.

src/main/java/uk/gov/register/presentation/SignedTreeHead.java
om-sharma: updated!!

PR #155, opened 2016-01-25
Rename of URL and/or parameters for /hash and /entry

AndrewGorton:
I can change /hash/{hash} to /item/{item-hash} and /entry/{serial-number} to /entry/{entry-number}, but /records should return entry/item xref but at present it returns entry+itemData.


philandstuff:
No redirect? :eyes:

philandstuff:
sorry, that was a little terse. what I mean to say is: did you have a conversation about whether to implement a redirect from the old URL to the new? If so, what was the outcome of that discussion?

PR #156, opened 2016-01-26
Get raw entries

AndrewGorton:
This gets the raw entries of the register as a zip file.
Each entry is a single JSON file in the items subdirectory.
Waiting on @michaelabroschomb for /register endpoint (for register.json file).
Entries and proof.json don't match the Sprint16 target yet also.


AndrewGorton:
For larger data sets, it might make more sense to generate the download archive periodically and stash it to a CDN location somewhere, rather than generating it on-the-fly.

PR #158, opened 2016-01-26
Creates a zipfile of all items

AndrewGorton:
This creates a zipfile on the fly containing all the register entries.

**For larger registers, this may not be the most efficient method.**

See http://openregister.github.io/specification/#download for a bit more details.

~~Depends upon the https://github.com/openregister/indexer/pull/35 for the timestamps in the Entry files, and probably makes sense to merge after https://github.com/openregister/presentation/pull/163.~~

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,6 +87,9 @@ test {
     testLogging {
         exceptionFormat = 'full'
     }
+    afterTest { desc, result ->
+        println "Executing test ${desc.name} [${desc.className}] with result: ${result.resultType}"
+    }
 }
 
 assemble {
--- a/src/main/java/uk/gov/register/presentation/Proofs.java
+++ b/src/main/java/uk/gov/register/presentation/Proofs.java
@@ -1,0 +1,15 @@
+package uk.gov.register.presentation;
+
+import uk.gov.register.proofs.ct.SignedTreeHead;
+
+public class Proofs {
+    private SignedTreeHead certificateTransparencySignedTreeHead;
+
+    public SignedTreeHead getCertificateTransparencySignedTreeHead() {
+        return certificateTransparencySignedTreeHead;
+    }
+
+    public void setCertificateTransparencySignedTreeHead(SignedTreeHead certificateTransparencySignedTreeHead) {
+        this.certificateTransparencySignedTreeHead = certificateTransparencySignedTreeHead;
+    }
+}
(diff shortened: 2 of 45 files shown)


src/main/java/uk/gov/register/presentation/Proofs.java
```diff
@@ -0,0 +1,15 @@
...
+import uk.gov.register.proofs.ct.SignedTreeHead;
+
+public class Proofs {
+    private SignedTreeHead certificateTransparencySignedTreeHead;
+
+    public SignedTreeHead getCertificateTransparencySignedTreeHead() {
+        return certificateTransparencySignedTreeHead;
+    }
+
+    public void setCertificateTransparencySignedTreeHead(SignedTreeHead certificateTransparencySignedTreeHead) {
```
philandstuff: our jackson data objects so far have used constructor injection rather than setter injection. could this class do the same for consistency?

src/main/java/uk/gov/register/presentation/resource/DataResource.java
```diff
@@ -1,41 +1,96 @@
...
-    public DataResource(ViewFactory viewFactory, RequestContext requestContext, RecentEntryIndexQueryDAO queryDAO) {
+    public DataResource(ViewFactory viewFactory, RequestContext requestContext, RecentEntryIndexQueryDAO queryDAO, SignedTreeHeadQueryDAO signedTreeHeadQueryDAO) {
         this.viewFactory = viewFactory;
         this.requestContext = requestContext;
         this.queryDAO = queryDAO;
+        this.signedTreeHeadQueryDAO = signedTreeHeadQueryDAO;
     }
 
     @GET
     @Path("/download")
```
philandstuff: comment to capture that we still need to confirm URL structure

philandstuff:
:+1: looks good so far but not merging per request

build.gradle
```diff
@@ -87,6 +87,9 @@ test {
     testLogging {
         exceptionFormat = 'full'
     }
+    afterTest { desc, result ->
+        println "Executing test ${desc.name} [${desc.className}] with result: ${result.resultType}"
```
AndrewGorton, line 91: For posterity: This is here so that I can check my test(s) actually ran. It emits the test name and result into the build log. Otherwise the only time I am sure my test ran is by deliberately making it fail.

I know that I could run gradle in INFO and DEBUG mode which would also display the test output, but it also fills the console with additional buildtime fluff.

PR #159, opened 2016-01-26
Make register domain configurable

philandstuff:
We currently hardcode the `openregister.org` domain in a bunch of different places in the code.  Since we're going to create a new set of registers on a new domain Real Soon Now, we should be able to make this configurable.

This PR finds all hardcoded uses of `openregister.org` and replaces them with a new configuration option instead.

There are a couple of talking points here:
1. To make the registerDomain configuration item available to the ThymeLeafView class, I just added it to the RequestContext class, but it clearly doesn't belong there. I'm happy to try to come up with a better solution, either before or after this is merged.
2. Prior to this PR, there are some places where we construct a hostname based on the hardcoded `openregister.org` suffix, and there are some places where we reuse the hostname that was in the given Host header. This PR only replaces the hardcoded usages with the configuration item, and leaves any place where we reuse the hostname from the Host header untouched.  This is perhaps clearest in the test data for turtle, eg single.ttl:

``` turtle
@prefix field: <http://field.test.register.gov.uk/field/>.

<http://register.beta.openregister.org/entry/1>
 field:fields <http://field.test.register.gov.uk/field/field1> ;
 field:register <http://register.test.register.gov.uk/register/value1> ;
 field:text "The Entry 1" .
```

the `register.beta.openregister.org` line is reused from the Host header, but all of the `test.register.gov.uk` lines are using the configuration item.  (In production, the Host header and the registerDomain header should agree with each other, but that won't always be the case.)

Hopefully these talking points shouldn't block this PR itself from being merged?

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -4,7 +4,7 @@ server:
     - type: http
       port: 9092
 
-registerDomain: .openregister.dev:8080
+registerDomain: openregister.dev:8080
 
 database:
   driverClass: org.postgresql.Driver
--- a/config.yaml
+++ b/config.yaml
@@ -4,6 +4,8 @@ server:
     - type: http
       port: 9092
 
+registerDomain: .openregister.dev:8080
+
 database:
   driverClass: org.postgresql.Driver
   url: jdbc:postgresql://localhost:5432/school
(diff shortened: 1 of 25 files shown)


om-sharma:
I like the PR description.  :+1:

PR #162, opened 2016-01-27
Create register resource for register.json

michaelabenyohai:
Added Total Entries to the HomePage for a register.

Created a /register resource which currently only returns json format. At some point we might want to merge the RegisterResourceView with the HomePageView if they are intended to display the same information. At the moment they are separate views.

Refactored the RegisterConfig and Register objects to retrieve the register entry as an EntryView, so that all representations of an entry remain consistent, whether just the entry or an entry nested in a register resource.

Still need to change last-updated property to return in sensible format. Also need to remove hardcoded registerDomain and reuse recent change made my @philandstuff to make registerDomain configurable.


philandstuff:
:fireworks:

PR #165, opened 2016-02-02
Return last-updated for register from leaf_input

michaelabenyohai:
This retrieves the leaf_input from the ordered_entry_index table for the row with the greatest serial_number. The getLastUpdated() method is changed to decode the leaf_input and return the timestamp. This timestamp is stored in the views as an Instant such that it is possible to display the UTC timezone when it is formatted as a date-time string in the json formats.

Also changed the RegisterResourceFunctionalTest to check that the timestamp is retrieved from the database correctly. The DBSupport class is changed to return the timestamp of the item added to the ordered_entry_index table with the greatest serial_number, such that it can be compared to the last-updated property of the /register.json returned.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/dao/RecentEntryIndexQueryDAO.java
+++ b/src/main/java/uk/gov/register/presentation/dao/RecentEntryIndexQueryDAO.java
@@ -11,9 +11,10 @@ import org.skife.jdbi.v2.sqlobject.customizers.RegisterMapper;
 import org.skife.jdbi.v2.sqlobject.customizers.SingleValueResult;
 import uk.gov.register.presentation.DbEntry;
 import uk.gov.register.presentation.mapper.EntryMapper;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/register/presentation/dao/RecentEntryIndexQueryDAO.java
```diff
@@ -40,6 +41,9 @@
     @SingleValueResult(DbEntry.class)
     public abstract Optional<DbEntry> findEntryBySerialNumber(@Bind("serial") long serial);
 
+    @SqlQuery("SELECT leaf_input FROM ordered_entry_index ORDER BY serial_number DESC LIMIT 1")
+    public abstract String getLatestTreeInput();
```
AndrewGorton: I think this is getLatestLeafInput().

PR #167, opened 2016-02-08
Display empty string when no leaf_input exists

om-sharma:
home page displays 500 error when no entries available in database. It was happening because leaf_input column has no value and we are extracting timestamp. 

Changes in PR displays empty string for timestamp when leaf_input value not present.


philandstuff:
:+1: happy to merge once you've updated the commit message

PR #175, opened 2016-02-15
Changed CSV line endings to LF only (previously CRLF).

AndrewGorton:
As mentioned in an email by @psd following a comment from users.


AndrewGorton:
Closed. Not needed at present.

PR #186, opened 2016-03-04
Remove usage of leaf_input.

om-sharma:
We are not going to use x-json sever (refer: https://github.com/openregister/mint/pull/64) which allows
us to remove usage of leaf_input column completely.

Proof stories are already being rewritten so removing leaf_input from register proof archive file won't have any issue.


michaelabenyohai:
Should CTLeafInputGenerator class also be removed? It doesn't appear to be used anywhere.

There is a test in HomePageFunctionalTest called homePageIsAvailableWhenLatestEntryHasNullLeafInput() - should this be removed?

The DBSupport class has a method called publishMessagesWithoutLeafInput, which probably does the same as publishMessages method now. Can this class be refactored to remove publishMessagesWithoutLeafInput?

om-sharma:
Removed!

PR #187, opened 2016-03-07
Added a caption to tables.

AndrewGorton:
Specifically added to entries, records, history and single entry tables. Functional tests added for all.


philandstuff:
@stephenjoe1 do these captions look okay to you?

PR #188, opened 2016-03-07
remove sth table

om-sharma:
Removed the sth table and all usage of it. [https://github.com/openregister/mint/pull/64]

Removed `/proof/certificate-transparency` resource also as no one is using it right now. Also the downloaded zip file of a register will not contain the proof.json now as new stories will follow for proof.


philandstuff:
Will this remove the register fingerprint from the front page? I'd like to keep that for the moment.

om-sharma:
No it doesn't. fingerprint on home page is hardcoded in html template which is kept intact.

src/test/java/uk/gov/register/presentation/functional/DataResourceTest.java
```diff
@@ -27,30 +29,24 @@ public static void publishTestMessages() {
...
-                if (entry.getName().compareTo("proof.json") == 0) {
-                    foundProofs = true;
-                } else if (entry.getName().compareTo("register.json") == 0) {
-                    foundRegister = true;
-                }
-                System.out.println(entry.getName());
-            }
-        } catch (IOException e) {
-            fail("Unexpected IO exception - " + e.getMessage());
+        assertTrue(getEntries(is).contains("register.json"));
```
philandstuff: can we stick to hamcrest please? Using hamcrest all the time results in consistent test failure reporting.  Also, when there's a failure you get something more than just "wanted true, got false" -- it will report what the list actually contained and not just what was being looked for.

``` json
assertThat(getEntries(is), hasItem("register.json"))
```

philandstuff:
:+1: other than that

src/test/java/uk/gov/register/presentation/functional/DataResourceTest.java
om-sharma: updated, check next commit.

PR #189, opened 2016-03-07
Redid the homepage links as specified in the trello card.

AndrewGorton:
Specifically, removed the link from the Last Updated field, and
added screenreader hidden elements into the records and entries
links (unfortunately, they're HTML-encoded in the source otherwise
Thymeleaf throws parsing exceptions).

### DIFF
--- a/src/main/resources/templates/home.html
+++ b/src/main/resources/templates/home.html
@@ -14,9 +14,9 @@
                 <dt>Provided by:</dt>
                 <dd><th:block th:include="main.html :: organisation(${custodian}, ${branding})"></th:block></dd>
                 <dt>Total records:</dt>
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/home.html
```diff
@@ -14,11 +14,11 @@
                 <dt>Provided by:</dt>
                 <dd><th:block th:include="main.html :: organisation(${custodian}, ${branding})"></th:block></dd>
                 <dt>Total records:</dt>
-                <dd><a href="/records" th:text="${totalRecords}">Total records</a></dd>
+                <dd><a href="/records" th:utext="${totalRecords + '&lt;span class=&quot;visuallyhidden&quot;&gt; records&lt;/span&gt;'}">Total records</a></dd>
```
philandstuff: What do you think of doing it this way?

``` html
<a href="/records"><th:block th:text="${totalRecords}">Total</th:block> <span class="visuallyhidden"> records</span></a>
```

this avoids the `th:utext` and the html-in-html thing.

PR #194, opened 2016-03-17
Serve same 404 for assets

AndrewGorton:
It's a bit ugly really, as I'm duplicating the Thymeleaf artifacts which we already create in ThymeleafView/Resolver/Renderer. 

An alternative approach would be to [use CloudFront to mask 404s](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/custom-error-pages.html).

Ideally you'd want to have the Thymeleaf artifacts injected into the object, but the HK2 ServiceLocator is null at the point of Application.run() and the jetty.setErrorHandler takes a class instance, not a class.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/AssetsBundleCustomErrorHandler.java
+++ b/src/main/java/uk/gov/register/presentation/AssetsBundleCustomErrorHandler.java
@@ -44,6 +44,10 @@ public class AssetsBundleCustomErrorHandler extends ErrorHandler {
 
     @Override
     public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException {
+        if (response.getStatus() != HttpServletResponse.SC_NOT_FOUND) {
+            return;
+        }
+
         baseRequest.setHandled(true);
 
         ServletContext sc = baseRequest.getContext();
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/register/presentation/AssetsBundleCustomErrorHandler.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+        templateResolver.setTemplateMode("HTML5");
+        templateResolver.setCacheable(false);
+        templateResolver.setCharacterEncoding(StandardCharsets.UTF_8.name());
+        engine = new TemplateEngine();
+        engine.setTemplateResolver(templateResolver);
+    }
+
+    @Override
+    public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException {
```
philandstuff, line 46: which errors does this handle? is it just 404s or all 4xxs or all 4xxs + all 5xxs?

AndrewGorton, line 46: Originally, it handled everything in AssetsBundle. Now I've added a check at the top which ensures it's only for 404s.

philandstuff:
looks like there's an encoding issue -- see "This is a new service ? your feedback"

screen shot 2016-03-17 at 14 51 53

Do you want to take a look at it before merging or shall I merge as is and you can look after?

AndrewGorton:
Lemme have a looksie. I was checking the /assets123 problem but that's mysteriously vanished instead.

PR #195, opened 2016-03-18
Display fields in the register register entry order

philandstuff:
Currently, we display fields in alphabetical order, possibly pulling the
primary key to the front.  This results in oddities such as pulling
`end-date` near the front and pushing `start-date` near the end.

The register register entry for a register, which defines which fields a
register has, happens to also have a reasonably sensible order of
fields.  We had been working on the assumption that this order is not
significant but it now seems that this order is nicer than any sorted
order would be.

In particular, every entry in the register register identifies the
primary field first before listing other fields.  I've kept our code
which pulls the primary key to the front for the moment but we could
perhaps replace it by just trusting the order in the register register.


philandstuff:
@psd we had an offhand discussion about this. what do you think? yay or nay?

philandstuff:
spoke with @stephenjoe1 and he's happy with this change.

PR #197, opened 2016-03-18
ignore all unknown fields in register/field config

philandstuff:
Register consumers must be [forwards-compatible](http://openregister.github.io/specification/#forwards-compatibility).  As part of this,
they must implement the must-ignore rule.  The best description I have
seen of this rule comes from the [I-JSON RFC](https://tools.ietf.org/html/rfc7493#section-4.2):

```
It is frequently the case that changes to protocols are required
after they have been put in production.  Protocols that allow the
introduction of new protocol elements in a way that does not disrupt
the operation of existing software have proven advantageous in
practice.

This can be referred to as a "Must-Ignore" policy, meaning that when
an implementation encounters a protocol element that it does not
recognize, it should treat the rest of the protocol transaction as if
the new element simply did not appear, and in particular, the
implementation MUST NOT treat this as an error condition.  The
converse "Must-Understand" policy does not tolerate the introduction
of new protocol elements, and while this has proven necessary in
certain protocol designs, in general it has been found to be overly
restrictive and brittle.

A good way to support the use of Must-Ignore in I-JSON protocol
designs is to require that top-level protocol elements must be JSON
objects, and to specify that members whose names are unrecognized
MUST be ignored.
```

Prior to this commit, we were not forwards-compatible against the
introduction of new fields in the register register or the field
register.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/RegisterData.java
+++ b/src/main/java/uk/gov/register/presentation/RegisterData.java
@@ -1,6 +1,7 @@
 package uk.gov.register.presentation;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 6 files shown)


philandstuff:
(yes this is basically the same as openregister/mint#70)

michaelabenyohai:
Should we use @JsonIgnoreProperties(ignoreUnknown = true) in PublicBody.java and PublicBodyConfiguration.java as well? At the moment those classes contain @JsonIgnoreProperties({"text","crest","website","parent-bodies","official-colour","public-body-type"}) and @JsonIgnoreProperties({"hash", "last-updated"}).

philandstuff:
good catch, yes, will do.

philandstuff:
updated

PR #198, opened 2016-03-21
Render copyright text as markdown

philandstuff:
The copyright field is defined to be markdown, and for the postcode
register the data is markdown.  This updates the copyright text to be
rendered as markdown.

Copyright text can appear in the main body or on the footer.  I've
written tests to catch both cases.

We are already rendering markdown on the front page from the register
`text` field, so this was relatively simple to add.

### DIFF
--- a/src/main/java/uk/gov/register/thymeleaf/ThymeleafView.java
+++ b/src/main/java/uk/gov/register/thymeleaf/ThymeleafView.java
@@ -17,7 +17,7 @@ import java.util.Optional;
 public class ThymeleafView extends View {
     protected final RequestContext requestContext;
     private String thymeleafTemplateName;
-    private final MarkdownProcessor markdownProcessor = new MarkdownProcessor();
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/register/thymeleaf/ThymeleafView.java
```diff
@@ -11,10 +12,12 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.nio.charset.StandardCharsets;
+import java.util.Optional;
 
 public class ThymeleafView extends View {
     protected final RequestContext requestContext;
     private String thymeleafTemplateName;
+    private final MarkdownProcessor markdownProcessor = new MarkdownProcessor();
```
om-sharma: Now markdownProcessor object is created here so I think we can use this object in HomePageView class also.

philandstuff: done

PR #201, opened 2016-03-23
Item resource

om-sharma:
Introducing new item resource.

It reads from new schema which might not be available in an environment, In such cases a 404 is returned. Once the schema with migrated data is available, the resource will start serving the content.

Currently EntryConverter is used to convert an item in to html displayable format, this inconsistency will be resolved once we start working on new Entry resource story.

### DIFF
--- a/create_schema.sh
+++ b/create_schema.sh
@@ -18,4 +18,8 @@ CREATE TABLE IF NOT EXISTS total_entries(count INTEGER, last_updated TIMESTAMP W
 
 insert into total_entries (count) select 0 where not exists (select count from total_entries);
 
+create table if not exists item (sha256hex varchar primary key, content jsonb);
+
+create table if not exists entry (entry_number integer primary key, sha256hex varchar, timestamp timestamp without time zone default now());
+
 "
--- a/drop_schema.sh
+++ b/drop_schema.sh
@@ -12,6 +12,10 @@ DROP TABLE IF EXISTS ordered_entry_index;
 
 DROP TABLE IF EXISTS total_entries;
 
+DROP TABLE IF EXISTS item;
+
+DROP TABLE IF EXISTS entry;
+
 DROP TABLE IF EXISTS sth; -- TODO: no longer used, should be deleted after a while
 
 "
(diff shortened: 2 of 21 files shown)


philandstuff:
I can't seem to view this resource locally? I get a 404 when I try:

screen shot 2016-03-23 at 12 51 59

but according to psql this hash is in the db:

```
$ psql school
psql (9.5.1)
Type "help" for help.

school=# \x
Expanded display is on.
school=# select * from item where sha256hex = '24cbed7ce7f57dd4786cfb2bb0338190daa576472f915e958b50ca612542b6b6';
-[ RECORD 1 ]----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
sha256hex | 24cbed7ce7f57dd4786cfb2bb0338190daa576472f915e958b50ca612542b6b6
content   | {"name": "Glanaman Home Tution Centre", "gender": "", "school": "402175", "address": "100101030506", "website": "", "end-date": "", "start-date": "2007-11-07", "headteacher": "  ", "maximum-age": "", "minimum-age": "", "religious-character": ""}
```

om-sharma:
Check resource url, it must be `/item/sha-256:<hex-value>` instead of `/item/sha256:<hex-value>`.

om-sharma:
@stephenjoe1: This is new item resource which is introduced as described in the specification. The html view for this is attached below. This will need some work, your comments?

screen shot 2016-03-23 at 13 34 32

philandstuff:
I think `immutable = true` is maybe a little aggressive for the html view here (though not the JSON view): if we want to change the page design, the old design might well still be cached. Maybe reduce to a day or a week?

om-sharma:
Discussed in person with phil and a new story is being created for caching policy across all cacheable resources in application.

philandstuff:
the old `/item` resource seems to be broken by this

om-sharma:
I assumed that there is test case which covers the old item resource, there was a test but it was not testing item resource at all. written new test and confirmed that it is working now.

PR #202, opened 2016-03-29
Entry resource

om-sharma:
Introducing new entry resource mentioned in http://openregister.github.io/specification/#entry-resource

### DIFF
--- a/src/main/java/uk/gov/register/presentation/dao/Entry.java
+++ b/src/main/java/uk/gov/register/presentation/dao/Entry.java
@@ -1,0 +1,24 @@
+package uk.gov.register.presentation.dao;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class Entry {
+    @JsonProperty("entry-number")
+    public final String entryNumber;
+
+    private final String sha256hex;
+
+    @JsonProperty("entry-timestamp")
+    public final String timestamp;
+
+    public Entry(String entryNumber, String sha256hex, String timestamp) {
+        this.entryNumber = entryNumber;
+        this.sha256hex = sha256hex;
+        this.timestamp = timestamp;
+    }
+
+    @JsonProperty("item-hash")
+    public String getSha256hex() {
+        return "sha-256:" + sha256hex;
+    }
+}
--- a/src/main/java/uk/gov/register/presentation/dao/Entry.java
+++ b/src/main/java/uk/gov/register/presentation/dao/Entry.java
@@ -2,16 +2,25 @@ package uk.gov.register.presentation.dao;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import java.sql.Timestamp;
+import java.text.SimpleDateFormat;
+import java.time.ZoneId;
+import java.util.TimeZone;
+
 public class Entry {
+    private final static SimpleDateFormat simpleDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
+    static {
+        simpleDateFormat.setTimeZone(TimeZone.getTimeZone(ZoneId.of("UTC")));
+    }
+
     @JsonProperty("entry-number")
     public final String entryNumber;
 
     private final String sha256hex;
 
-    @JsonProperty("entry-timestamp")
-    public final String timestamp;
+    private final Timestamp timestamp;
 
-    public Entry(String entryNumber, String sha256hex, String timestamp) {
+    public Entry(String entryNumber, String sha256hex, Timestamp timestamp) {
         this.entryNumber = entryNumber;
         this.sha256hex = sha256hex;
         this.timestamp = timestamp;
@@ -20,5 +29,10 @@ public class Entry {
     @JsonProperty("item-hash")
     public String getSha256hex() {
         return "sha-256:" + sha256hex;
+    }
+
+    @JsonProperty("entry-timestamp")
+    public String getTimestamp() {
+        return simpleDateFormat.format(timestamp);
     }
 }
--- a/src/main/java/uk/gov/register/presentation/resource/EntryResource.java
+++ b/src/main/java/uk/gov/register/presentation/resource/EntryResource.java
@@ -24,6 +24,7 @@ public class EntryResource {
     private final ViewFactory viewFactory;
     private final RecentEntryIndexQueryDAO queryDAO;
     private final RequestContext requestContext;
+    private static final String POSTGRES_TABLE_NOT_EXIST_ERROR_CODE = "42P01";
 
 
     @Inject
@@ -43,7 +44,7 @@ public class EntryResource {
             Optional<Entry> entry = entryDAO.findByEntryNumber(entryNumber);
             return entry.map(viewFactory::getNewEntryView).orElseThrow(NotFoundException::new);
         } catch (Throwable e) {
-            if (e.getCause() instanceof PSQLException && ((PSQLException) e.getCause()).getSQLState().equals("42P01")) {
+            if (e.getCause() instanceof PSQLException && ((PSQLException) e.getCause()).getSQLState().equals(POSTGRES_TABLE_NOT_EXIST_ERROR_CODE)) {
                 return findBySerial(Optional.of(entryNumber));
             }
             throw e;
--- a/src/main/java/uk/gov/register/presentation/resource/EntryResource.java
+++ b/src/main/java/uk/gov/register/presentation/resource/EntryResource.java
@@ -1,35 +1,68 @@
 package uk.gov.register.presentation.resource;
 
 import io.dropwizard.jersey.caching.CacheControl;
+import org.postgresql.util.PSQLException;
+import uk.gov.register.presentation.DbEntry;
 import uk.gov.register.presentation.dao.Entry;
 import uk.gov.register.presentation.dao.EntryDAO;
+import uk.gov.register.presentation.dao.RecentEntryIndexQueryDAO;
 import uk.gov.register.presentation.representations.ExtraMediaType;
-import uk.gov.register.presentation.view.NewEntryView;
+import uk.gov.register.presentation.view.AttributionView;
+import uk.gov.register.presentation.view.SingleEntryView;
 import uk.gov.register.presentation.view.ViewFactory;
 
 import javax.inject.Inject;
 import javax.ws.rs.*;
 import javax.ws.rs.core.MediaType;
 import java.util.Optional;
+import java.util.function.Function;
 
 @Path("/entry")
 public class EntryResource {
 
     private final EntryDAO entryDAO;
     private final ViewFactory viewFactory;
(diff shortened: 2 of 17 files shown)


src/main/java/uk/gov/register/presentation/dao/Entry.java
```diff
@@ -0,0 +1,38 @@
+package uk.gov.register.presentation.dao;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.sql.Timestamp;
```
philandstuff, line 5: this is a mutable object class. is there an immutable alternative? Perhaps from joda?

philandstuff, line 5: looks like dropwizard-jdbi adds joda time support: https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jdbi/src/test/java/io/dropwizard/jdbi/timestamps/JodaDateTimeSqlTimestampTest.java if this is easy I'd prefer the joda types rather than the java.sql types.

src/main/java/uk/gov/register/presentation/dao/Entry.java
```diff
@@ -0,0 +1,38 @@
...
+import java.text.SimpleDateFormat;
+import java.time.ZoneId;
+import java.util.TimeZone;
+
+public class Entry {
+    private final static SimpleDateFormat simpleDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
```
philandstuff, line 11: is there an existing formatter for this? given we're going for standard RFC3339 I'd expect there to be an existing formatter

src/main/java/uk/gov/register/presentation/resource/EntryResource.java
```diff
@@ -0,0 +1,68 @@
...
+    @GET
+    @Path("/{entry-number}")
+    @Produces({ExtraMediaType.TEXT_HTML, MediaType.APPLICATION_JSON})
+    @CacheControl(immutable = true)
+    public AttributionView findByEntryNumber(@PathParam("entry-number") int entryNumber) {
+        try {
+            Optional<Entry> entry = entryDAO.findByEntryNumber(entryNumber);
+            return entry.map(viewFactory::getNewEntryView).orElseThrow(NotFoundException::new);
+        } catch (Throwable e) {
+            if (e.getCause() instanceof PSQLException && ((PSQLException) e.getCause()).getSQLState().equals("42P01")) {
```
philandstuff: `"42P01"` can this magic number be extracted as a constant?

philandstuff: also, why is this special case here at all?

src/main/java/uk/gov/register/presentation/resource/EntryResource.java
```diff
@@ -0,0 +1,68 @@
...
+    public EntryResource(EntryDAO entryDAO, ViewFactory viewFactory, RecentEntryIndexQueryDAO queryDAO, RequestContext requestContext) {
+        this.entryDAO = entryDAO;
+        this.viewFactory = viewFactory;
+        this.queryDAO = queryDAO;
+        this.requestContext = requestContext;
+    }
+
+    @GET
+    @Path("/{entry-number}")
+    @Produces({ExtraMediaType.TEXT_HTML, MediaType.APPLICATION_JSON})
```
philandstuff, line 40: do we need to write another card for other representations here?

src/test/java/uk/gov/register/presentation/functional/EntryResourceFunctionalTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Test
+    public void getEntryByEntryNumber() throws JSONException, IOException {
+        String sha256Hex = DigestUtils.sha256Hex(item1);
+
+        Response response = getRequest("/entry/1.json");
+
+        assertThat(response.getStatus(), equalTo(200));
+        assertThat(response.getHeaders().get("cache-control").toString(), equalTo("[no-transform, max-age=31536000]"));
```
philandstuff: why are the square brackets here? I would expect a valid cache-control header to look like this:

```
Cache-Control: no-transform, max-age=31536000
```

src/main/java/uk/gov/register/presentation/dao/Entry.java
om-sharma, line 5: I agree we must not use Timestamp class for this purpose. java8 got new DateTime API which is immutable and much improved version, can use that.

src/main/java/uk/gov/register/presentation/dao/Entry.java
om-sharma, line 11: This will be changed if we don't use `java.sql.Timestamp` class.

src/test/java/uk/gov/register/presentation/functional/EntryResourceFunctionalTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+        JsonNode res = Jackson.newObjectMapper().readValue(response.readEntity(String.class), JsonNode.class);
+        assertThat(res.get("entry-number").textValue(), equalTo("1"));
+        assertThat(res.get("item-hash").textValue(), equalTo("sha-256:" + sha256Hex));
+        assertThat(res.get("entry-timestamp").textValue(), containsString(LocalDate.now().toString()));
```
philandstuff: this looks likely to become a flaky test due to its timing dependence. Maybe we could assert the timestamp is "some time in the last minute"? :confused:

src/test/java/uk/gov/register/presentation/functional/RepresentationsTest.java
```diff
@@ -3,6 +3,7 @@
 import com.google.common.base.Charsets;
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.Resources;
+import jdk.nashorn.internal.ir.annotations.Ignore;
```
philandstuff: don't think this is the right annotation?

src/test/java/uk/gov/register/presentation/functional/RepresentationsTest.java
```diff
@@ -50,7 +51,7 @@ public RepresentationsTest(String extension, String expectedContentType, String
         this.expectedListOfEntries = expectedListOfEntries;
     }
 
-    @Test
```
philandstuff, line 53: you don't need to remove `@Test` when adding `@Ignore` IIRC

src/test/java/uk/gov/register/presentation/functional/RepresentationsTest.java
```diff
@@ -50,7 +51,7 @@ public RepresentationsTest(String extension, String expectedContentType, String
...
+    @Ignore
```
philandstuff, line 54: what is your plan for this test? How will it become un-ignored again?

src/main/java/uk/gov/register/presentation/resource/EntryResource.java
om-sharma: `42P01` is postgres error code which is returned in case of queried table does not exist. There is a unit test case to explain this, we can also extract it as a constant. 
Special case is required so that we don't have deployment dependencies between indexer and presentation. If new schema is not available the old resource will respond otherwise the new resource response will be returned.

src/main/java/uk/gov/register/presentation/resource/EntryResource.java
om-sharma, line 40: I assumed the current representation card covers all the new resource. we can have either on card or one for each new resource.

src/test/java/uk/gov/register/presentation/functional/RepresentationsTest.java
om-sharma: Yes, You are absolutely correct. Its because I am a nut, rely a lot on IDE and don't look back the changes.

src/test/java/uk/gov/register/presentation/functional/RepresentationsTest.java
om-sharma, line 54: I didn't want to do the hard work again very soon when we write other representations for new resources, so kept the test. When we play the representations card this test will be unignored.

src/test/java/uk/gov/register/presentation/functional/RepresentationsTest.java
om-sharma, line 53: Yes, You are correct.

src/test/java/uk/gov/register/presentation/functional/EntryResourceFunctionalTest.java
om-sharma: Brackets are because header value is represented in a list. I am ok with this but if you like I can write more verbose assert in here.

src/test/java/uk/gov/register/presentation/functional/EntryResourceFunctionalTest.java
om-sharma: I agree it can fail in some rare conditions, I was in two minds of either matching the timestamp format or some value in it. Now changed it to match the timestamp format.

src/test/java/uk/gov/register/presentation/functional/EntryResourceFunctionalTest.java
philandstuff: If it's a list value, can you use the `contains()` or
`containsInAnyOrder()` matchers?

om-sharma: `contains()` does not confirm that the list contains only asserted values, size check will also be required.

philandstuff: [you would think so, wouldn't you?](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#contains%28E...%29) but no, it checks for _exact_ contents.

om-sharma: Ok, didn't know that. changed it.

src/main/java/uk/gov/register/presentation/resource/ThrowableExceptionMapper.java
```diff
@@ -30,6 +31,12 @@ public Response toResponse(Throwable exception) {
             throw (ClientErrorException) exception;
         }
 
+        if (exception instanceof ParamException.PathParamException) {
```
philandstuff, line 34: what can cause this? do we need a test for it?

om-sharma, line 34: There is a functional test for this `EntryResourceFunctionalTest.return404ResponseWhenEntryNumberIsNotAnIntegerValue()`.

philandstuff, line 34: :+1:

PR #204, opened 2016-03-30
Entries resource

om-sharma:
Note: Please do not merge this PR before https://github.com/openregister/presentation/pull/202. this will need rebase once PR #202 is merged.

This PR introduces new `/entries` resource to match the specification response. In case the old schema is not migrated the old `/entries` resource is served.

screen shot 2016-04-06 at 15 23 33


om-sharma:
Merge this PR only after #202

src/main/java/uk/gov/register/presentation/dao/EntryDAO.java
```diff
@@ -17,6 +18,10 @@
     @SingleValueResult(Entry.class)
     Optional<Entry> findByEntryNumber(@Bind("entry_number") int entryNumber);
 
+    @SqlQuery("select * from entry limit 100")
+    @RegisterMapper(NewEntryMapper.class)
+    List<Entry> getAll();
```
philandstuff: not sure `getAll` is the right name for this method

om-sharma: why do you think so? is it because of there is a limit of 100 entries in query? 
Any other name preference? I can rename it to like `getEntries`.

om-sharma:
Renamed the method name to `getEntries`.

om-sharma:
Added pagination for new `/entries` resource

PR #206, opened 2016-04-06
New Record resource

om-sharma:
Merge only after #204


om-sharma:
Closing because we want the target branch master and there is no easy way to update the target branch.

PR #207, opened 2016-04-07
Introducing `/record/<key>` resource

om-sharma:


philandstuff:
Looks like we need a card for the "record entries" ("history") resource?

om-sharma:
added a card for this.

PR #208, opened 2016-04-07
Render json format as only other available data representation for new resources.

om-sharma:
New resource only has Json representation available other than html, so available data-formats in footer only shows json.


src/main/resources/templates/item.html
```diff
@@ -8,13 +8,14 @@
     <div th:replace="main.html::phase"></div>
     <div class="grid-row">
         <div class="column-two-thirds">
+            <p></p>
```
om-sharma, line 11: added this empty `<p>` tag temporarily to right align the attribution so that it is consistent with other pages.

PR #209, opened 2016-04-07
Implement yaml representation for new entry resource

philandstuff:
As part of this change I've updated the DBSupport infrastructure to
allow setting explicit entry timestamps in test data.

### DIFF
--- a/src/test/java/uk/gov/register/presentation/functional/testSupport/TestDAO.java
+++ b/src/test/java/uk/gov/register/presentation/functional/testSupport/TestDAO.java
@@ -1,7 +1,10 @@
 package uk.gov.register.presentation.functional.testSupport;
 
+import io.dropwizard.java8.jdbi.args.InstantArgumentFactory;
 import org.skife.jdbi.v2.DBI;
 import org.skife.jdbi.v2.Handle;
+
+import java.util.Optional;
 
 public class TestDAO {
     public final TestOrderedEntryIndexDAO testEntryIndexDAO;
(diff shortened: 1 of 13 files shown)


src/test/resources/fixtures/single.yaml
```diff
@@ -1,8 +1,4 @@
 ---
-serial-number: 1
-hash: "someHash1"
-entry:
-  fields:
-  - "field1"
-  register: "value1"
-  text: "The Entry 1"
+entry-number: "1"
```
philandstuff, line 2: Hold on a minute, I haven't worked out why this turned into a string.

om-sharma, line 2: Specification https://openregister.github.io/specification/#entry-resource specifies all field values as string.

philandstuff, line 2: ah okay, and https://openregister.github.io/specification/#json-representation says

> All field values MUST be encoded as JSON strings.

for the JSON representation. We should probably decide if we want to do the same for YAML or not.

philandstuff, line 2: I raised openregister/specification#51 for this issue. Let's just go with strings for the moment, we can change it later if necessary.

src/test/java/uk/gov/register/presentation/functional/testSupport/TestDAO.java
```diff
@@ -15,6 +18,7 @@
     private TestDAO(String databaseName, String user) {
         String postgresConnectionString = String.format("jdbc:postgresql://localhost:5432/%s?user=%s", databaseName, user);
         DBI dbi = new DBI(postgresConnectionString);
+        dbi.registerArgumentFactory(new InstantArgumentFactory(Optional.empty()));
```
om-sharma, line 21: There is already a default constructor of InstantArgumentFactory.

PR #210, opened 2016-04-08
Item tests

philandstuff:
Adds YAML representation for the /item resource


src/main/java/uk/gov/register/presentation/representations/NewRepresentationWriter.java
```diff
@@ -27,15 +30,21 @@ public final long getSize(View t, Class<?> type, Type genericType, Annotation[]
...
+        return NewEntryView.class.isAssignableFrom(type) ||
+                ItemView.class.isAssignableFrom(type);
     }
 
     @Override
     public void writeTo(View view, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream) throws IOException, WebApplicationException {
         if (view instanceof NewEntryView) {
             writeEntryTo(entityStream, requestContext.getRegister().getFields(), ((NewEntryView) view).getEntry());
         }
+        if (view instanceof ItemView) {
```
om-sharma: Is it possible to implement `writeTo` method in such a way that it can convert the view object in any representation without knowing its type?

src/main/java/uk/gov/register/presentation/representations/NewRepresentationWriter.java
```diff
@@ -27,15 +30,21 @@ public final long getSize(View t, Class<?> type, Type genericType, Annotation[]
...
+            writeItemTo(entityStream, ((ItemView) view).getContent());
+        }
```
philandstuff: @om-sharma raised concerns IRL that this `if (foo instanceof Bar)` chain is going to result in ugly code as we add more resources to the product.  He suggested instead that, given we already have a way of producing JSON for each resource, in principle all we should need to have is a generic way of producing another representation from JSON and then we don't need this special casing.

I'm not sure I agree but I'll investigate and I wanted to capture this context on the PR. I think we both agreed that this approach might not work for the turtle type.

philandstuff:
I've updated the PR to go with a simplified NewRepresentationWriter which we can work with for the moment. If it turns out that it doesn't work with CSV/TSV/turtle we can always change it again later.

PR #217, opened 2016-04-13
csv representation for new entry, entries, and item resources

michaelabenyohai:
The new views now implement an interface called RepresentationView and it's only method, csvSchema(). The NewCsvWriter then uses this method to obtain the CsvSchema for that particular resource.


src/main/java/uk/gov/register/presentation/view/ItemView.java
```diff
@@ -26,4 +28,13 @@
     public Map<String, FieldValue> getContent() {
         return item.getFieldsStream().collect(Collectors.toMap(Map.Entry::getKey, itemConverter::convert));
     }
+
+    @Override
+    public CsvSchema csvSchema() {
+        CsvSchema.Builder schemaBuilder = new CsvSchema.Builder() ;
+        for (String value : getContent().keySet()) {
```
philandstuff: Hmm, thinking about this I'm not sure it's quite right.

It might work for the `/item` resource at the moment, but for the `/records` (or later `/items`) resource, we need a single schema that will work for arbitrary records (or items).  This means we need to include all possible fields in the schema, even fields which aren't present in a particular item.

So, for example, if a `country` doesn't have an `end-date`, I think we should still have an `end-date` column in the CsvSchema.

src/test/resources/fixtures/item.csv
```diff
@@ -0,0 +1,2 @@
+text,fields,register
+"The Entry 1",field1,value1
```
philandstuff: do we need a test for an item containing multiple `fields`?  Previously we had a test to ensure we could see that JSON `"fields":["field1","field2"]` was rendered as a CSV column `field1;field2`; this was in the `/records` resource test which might explain why we've lost it here.

philandstuff: oops, just saw `writeEntriesTo_writesLists` above already covers this case. never mind, move along... 😄

src/main/java/uk/gov/register/presentation/view/ItemView.java
```diff
@@ -26,4 +28,13 @@
     public Map<String, FieldValue> getContent() {
         return item.getFieldsStream().collect(Collectors.toMap(Map.Entry::getKey, itemConverter::convert));
     }
+
+    @Override
+    public CsvSchema csvSchema() {
+        CsvSchema.Builder schemaBuilder = new CsvSchema.Builder() ;
+        for (String value : getRegister().getFields()) {
```
michaelabenyohai, line 35: @philandstuff this is now using register fields as headers in building the schema

PR #222, opened 2016-04-18
Implement tsv representation for new resources

michaelabenyohai:
Including entries, entry, item, record and record entries resources


src/main/resources/templates/data-formats.html
```diff
@@ -13,8 +13,10 @@
 <div th:fragment="new-data-formats">
     <p class="data-representations">This data is available as:
         <a th:href="${@uk.gov.register.presentation.HtmlViewSupport@representationLink(#httpServletRequest, 'json')}">json</a>,
-        <a th:href="${@uk.gov.register.presentation.HtmlViewSupport@representationLink(#httpServletRequest, 'yaml')}">yaml</a> and
-        <a th:href="${@uk.gov.register.presentation.HtmlViewSupport@representationLink(#httpServletRequest, 'csv')}">csv</a>
+        <a th:href="${@uk.gov.register.presentation.HtmlViewSupport@representationLink(#httpServletRequest, 'yaml')}">yaml</a>,
+        <a th:href="${@uk.gov.register.presentation.HtmlViewSupport@representationLink(#httpServletRequest, 'html')}">html</a>,
```
philandstuff, line 17: not sure we need an explicit link to html? think this is just confusing, particularly since if you're reading this text you've already got the html 😄

PR #227, opened 2016-04-20
Restructure DAO methods in respective DAO classes

om-sharma:
Moving DAO methods from old schema `RecentEntryIndexQueryDAO`  class to new DAO classes so that it is easy to understand about which methods are actually used and then gradually remove their usage and finally delete the old schema.


philandstuff:
👍

PR #230, opened 2016-04-22
Update download resource to use new schema

om-sharma:
Also Refactor ArchiveCreator to make it more readable

### DIFF
--- a/src/main/java/uk/gov/register/presentation/ArchiveCreator.java
+++ b/src/main/java/uk/gov/register/presentation/ArchiveCreator.java
@@ -11,7 +11,6 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.List;
-import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/register/presentation/ArchiveCreator.java
```diff
@@ -1,77 +1,78 @@
...
         return output -> {
-            ZipOutputStream zos = new ZipOutputStream(output);
+            try (ZipEntryWriter zipEntryWriter = new ZipEntryWriter(output)) {
+
+                zipEntryWriter.writeEntry("register.json", registerDetail);
 
-            ObjectMapper om = new ObjectMapper();
-            om.configure(JsonGenerator.Feature.AUTO_CLOSE_TARGET, false);
+                records.stream()
+                        .collect(Collectors.toMap(e -> e.entry.getSha256hex(), e -> e))
```
michaelabenyohai: If there are any entries in this list of records that refer to the same item, this will raise a "duplicate key" exception since e.entry.getSha256hex() will be the same. Should the create() method take separate lists of Entry and Item objects to avoid this?

om-sharma: Good catch!!
I have updated the test and actual code.

PR #231, opened 2016-04-22
Migrate record link

om-sharma:
We have old records resource which is still available because of e-petitions is using it. all records on this page are now linked to new record resource which allows us to remove the old item resource and old history resource because these will not be available by any path now. It also reduces the dependency over old schema.

check individual commits for more details.


src/main/java/uk/gov/register/presentation/resource/RegisterResource.java
```diff
@@ -0,0 +1,38 @@
...
+    @Inject
+    public RegisterResource(ViewFactory viewFactory, RecordDAO recordDAO, EntryDAO entryDAO) {
+        this.viewFactory = viewFactory;
+        this.recordDAO = recordDAO;
+        this.entryDAO = entryDAO;
+    }
+
+    @GET
+    @Path("/register")
+    @Produces({MediaType.APPLICATION_JSON})
```
philandstuff, line 29: should produce more than just JSON?

philandstuff, line 29: huh, I guess this is just reproducing the old behaviour. /shrug

philandstuff:
There are still some links to the old records resource:
- The rendered fields on `/record` and `/item` pages
- The field links (ie links to the field register when you click on a field name)
- "View this entry in the register of registers"

These are all fine because of the redirect from old record resource to new record resource, but they will need cleaning up.

PR #233, opened 2016-04-22
Replace old records resource from new schema resource

om-sharma:
review separate commits


philandstuff:
Are we confident that alphagov/e-petitions#455 has been deployed to prod?

om-sharma:
Not sure about it. Will ask mike.seipp to confirm it.

michaelabenyohai:
This also now needs to be rebased.

PR #241, opened 2016-04-28
New resource ttl representation

michaelabenyohai:
All Views that extend RepresentationView now have to implement a turtleRepresentation() method, which returns a Jena Model object. When the NewTurtleWriter is called it obtains the Model for the View and writes to Turtle.


src/main/java/uk/gov/register/presentation/representations/NewTurtleWriter.java
```diff
@@ -0,0 +1,24 @@
...
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+@Provider
+@Produces(ExtraMediaType.TEXT_TTL)
+public class NewTurtleWriter extends NewRepresentationWriter {
+    
+    public NewTurtleWriter() {
+    }
```
philandstuff: very minor thing, but I don't think this constructor is needed

src/main/java/uk/gov/register/presentation/view/RegisterDetailView.java
```diff
@@ -49,4 +50,9 @@ public RegisterDetail getRegisterDetail() {
     public CsvRepresentation<RegisterDetail> csvRepresentation() {
         throw new NotImplementedException("Not implemented");
     }
+
+    @Override
+    public Model turtleRepresentation() {
+        throw new NotImplementedException("Not implented");
```
philandstuff: why are we implementing RepresentaionView if we don't want to actually implement it?

src/test/resources/fixtures/list.ttl
```diff
@@ -1,11 +1,24 @@
...
+      field:text "The Entry 1" ;
+      register-metadata:entry-number-field
+              "1" ;
+      register-metadata:entry-timestamp-field
+              "2016-03-01T01:02:03Z" ;
+      register-metadata:item-resource
+              <http://register.test.register.gov.uk/item/sha-256:877d8bd1ab71dc6e48f64b4ca83c6d7bf645a1eb56b34d50fa8a833e1101eb18> .
+
+<http://register.test.register.gov.uk/record/value2>
+      field:fields field:field1 , field:field2 ;
```
philandstuff, line 16: huh, I didn't know you could do lists in turtle like this

philandstuff:
looks good on first glance. I'll try to give it a more thorough review after lunch.

I'm a little concerned that this makes our view classes quite bloated (eg ItemView) with code that's specific only to RDF stuff. I'd be interested in having a conversation about we might avoid this (and whether it's worth it)

michaelabenyohai:
@philandstuff - latest commits should address your review comments.

philandstuff:
🐢

PR #243, opened 2016-05-03
Cleanup dbsupport

om-sharma:
Old schema cleanup and also made ignored test active


src/test/java/uk/gov/register/presentation/functional/TestEntry.java
```diff
@@ -7,9 +7,17 @@
...
 
-    public TestEntry(int entryNumber, String itemJson, Instant entryTimestamp) {
+    private TestEntry(int entryNumber, String itemJson, Instant entryTimestamp) {
         this.entryNumber = entryNumber;
         this.itemJson = itemJson;
         this.entryTimestamp = entryTimestamp;
     }
+
+    public static TestEntry anEntry(int entryNumber, String itemJson){
+        return new TestEntry(entryNumber, itemJson, Instant.now());
```
philandstuff, line 33: should this be `now()` or could it be a fixed value instead? I'm just thinking at a high level about how we can avoid potential sources of test flakiness.

src/test/java/uk/gov/register/presentation/functional/testSupport/DBSupport.java
```diff
@@ -13,19 +11,20 @@
...
+
 public class DBSupport {
     public static void main(String[] args) throws IOException, SQLException {
         String filePath = args[1];
 
         DBSupport dbSupport = new DBSupport(TestDAO.get(args[2], "postgres"));
         try (Stream<String> lines = Files.lines(new File(filePath).toPath(), Charset.defaultCharset())) {
-            List<String> collect = lines.collect(Collectors.toList());
-            dbSupport.publishMessages(args[0], collect);
+            AtomicInteger index = new AtomicInteger(0);
```
philandstuff, line 23: creative use of AtomicInteger, though it feels a little dodgy because it depends on the exact order of processing of the `map()` call. It's a shame there's no `mapIndexed()` function :/

src/test/java/uk/gov/register/presentation/functional/testSupport/TestEntryDAO.java
```diff
@@ -12,9 +12,6 @@
     @SqlUpdate("create table if not exists entry (entry_number integer primary key, sha256hex varchar, timestamp timestamp without time zone default now())")
```
philandstuff: can/should we remove the `default now()` bit here?

src/test/java/uk/gov/register/presentation/functional/TestEntry.java
```diff
@@ -1,23 +1,49 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.google.common.base.Throwables;
+import org.apache.commons.codec.digest.DigestUtils;
+
+import java.io.IOException;
 import java.time.Instant;
 
 public class TestEntry {
+    private static final ObjectMapper objectMapper = new ObjectMapper();
```
philandstuff: can/should this use `PresentationApplication.customObjectMapper()`?

philandstuff:
👍 great to see this change. Let's discuss IRL and hopefully get it merged asap

src/test/java/uk/gov/register/presentation/functional/TestEntry.java
om-sharma, line 33: There are two methods. The other method takes `instant` value as a parameter. Tests which are interested in `entry-timestamp` value can use the other method.

src/test/java/uk/gov/register/presentation/functional/testSupport/DBSupport.java
om-sharma, line 23: Using `map` method on a list stream shouldn't be a problem as it keeps the ordering (if I understood your concern correctly).

This is used in `main` method to load the data for development environment. An entry number  can be provided with in the data file instead of using `AtomicInteger` iff needed.

src/test/java/uk/gov/register/presentation/functional/testSupport/TestEntryDAO.java
om-sharma: Yes can remove. will update in next commit.

src/test/java/uk/gov/register/presentation/functional/TestEntry.java
om-sharma: No, I strongly think that we must not couple PresentationApplication class in here or anywhere. I agree we can extract `customObjectMapper()` in separate class and can reuse if required.

Also this ObjectMapper is configured with serialization feature `SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS` and customObjectMapper is not.

PR #247, opened 2016-05-13
New Pagination support for `/entries` resource

om-sharma:
This introduces new pagination support for `/entries` resource. 
Please check trello card to understand more about the story.


src/test/java/uk/gov/register/presentation/resource/NewPaginationTest.java
```diff
@@ -0,0 +1,103 @@
...
+    @Test
+    public void getLastEntryNumberOnThisPage_returnsTheLastEntryNumber() {
+        assertThat(new NewPagination(Optional.of(11), Optional.of(100), 90).getLastEntryNumberOnThisPage(), equalTo(90));
+        assertThat(new NewPagination(Optional.of(11), Optional.of(100), 120).getLastEntryNumberOnThisPage(), equalTo(110));
+    }
+
+    @Test
+    public void getPreviousPageNumber_returnsThePreviousPageNumber() {
+        assertThat(new NewPagination(Optional.of(11), Optional.of(100), 100).getPreviousPageNumber(), equalTo(1));
+        assertThat(new NewPagination(Optional.of(11), Optional.of(100), 100).getPreviousPageNumber(), equalTo(1));
```
michaelabenyohai: This test appears to be duplicated.

src/main/java/uk/gov/register/presentation/view/EntryListView.java
```diff
@@ -5,34 +5,43 @@
...
+    public Collection<Entry> getEntries() {
         return entries;
     }
 
     @SuppressWarnings("unused, used from templates")
-    public Pagination getPagination() {
-        return pagination;
+    public NewPagination getPagination() {
+        return newPagination;
     }
```
michaelabenyohai, line 33: This change breaks the /record/xxxx/entries resource. If you go to that resource it produces a template error because the newPagination property is null (the pagination property is set instead for this case and never returned to the view).

Suggest that EntryListView is changed to use single pagination property and constructor, either:
- Pagination and NewPagination classes should implement a single interface called IPagination, for example. Then getPagination() should return this and there should be one constructor that takes an IPagination.
- or it should only take a NewPagination property for both resources and the getAllEntriesOfARecord() method in the recordResource should create a hardcoded NewPagination object similarly to how it does currently with a Pagination object.

src/test/java/uk/gov/register/presentation/resource/NewPaginationTest.java
om-sharma: That's true, I shall remove this.

src/main/java/uk/gov/register/presentation/view/EntryListView.java
om-sharma, line 33: I would like to remove the Old pagination object so didn't do anymore work. Have you had any discussion about the questions above?

src/main/java/uk/gov/register/presentation/view/ViewFactory.java
```diff
@@ -65,7 +67,11 @@ public EntryView getEntryView(Entry entry) {
         return new EntryView(requestContext, getCustodian(), getBranding(), entry);
     }
 
-    public EntryListView getEntryListView(List<Entry> entries, Pagination pagination) {
+    public EntryListView getEntryListView(List<Entry> entries, IPagination pagination) {
+        return new EntryListView(requestContext, pagination, getCustodian(), getBranding(), entries);
+    }
+
```
michaelabenyohai: Why do we need this method now when we also have getEntriesView, which is the same except that it takes a Collection instead of a List? Can we use getEntriesView in both places and remove getEntryListView?

src/main/java/uk/gov/register/presentation/resource/NewPagination.java
```diff
@@ -14,28 +16,38 @@ public NewPagination(Optional<Integer> optionalStart, Optional<Integer> optional
...
+    @Override
     public String getPreviousPageLink() {
         return String.format("?start=%s&limit=%s", start - limit, limit);
     }
 
+    @Override
+    public long pageSize() {
+        throw new NotImplementedException("not implemented");
+    }
+
```
michaelabenyohai: Can we remove pageSize() from IPagination so that we don't need to override it and throw a NotImplementedException?

om-sharma:
@michaelabenyohai: updated

PR #255, opened 2016-07-01
RFC: Make scheme explicitly configurable (default https)

philandstuff:
**Please don't merge this yet -- at minimum, we'll need to change all the app config files first**

Currently, we attempt to sniff the register scheme from the request
context.  This fails because the request at the application server may
not use the same scheme as the request at the edge server.  For example,
the app server might be running http, but the edge server might be a CDN
terminating https.

The consequences of this are that absolute URLs are always `http`
urls. In beta, these get automatically converted to `https` (either by
301 redirect or by HSTS in the browser).  These URLs appear both as href
attributes in html links, and as entities and attributes within the
turtle representation.

There are two alternatives to fix this:
- use the X-Forwarded-Proto header to determine what the URI scheme at
  the edge server was
- make the scheme an explicit configuration option

I think the Right Thing is to use X-Forwarded-Proto but that's a bit
more work.  The Easy Thing is to make the scheme an explicit
configuration option and set the config up everywhere.  We can do the
Easy Thing now in order to get correct behaviour, and refactor to the
Right Thing later.

There's a third option which is to use https everywhere -- including in
tests, and when starting it up locally or in a heroku instance or
whatever. I'm not sure I want to go down that road.

### DIFF
--- a/src/test/resources/test-app-config.yaml
+++ b/src/test/resources/test-app-config.yaml
@@ -8,6 +8,7 @@ server:
       port: 9001
 
 registerDomain: test.register.gov.uk
+registerScheme: http
 
 database:
   driverClass: org.postgresql.Driver
--- a/src/main/java/uk/gov/register/presentation/HtmlViewSupport.java
+++ b/src/main/java/uk/gov/register/presentation/HtmlViewSupport.java
@@ -19,7 +19,7 @@ public class HtmlViewSupport {
         return uriBuilder.build().toString();
     }
 
-    public static String fieldLink(String fieldName, String registerDomain, String requestScheme) {
-        return new LinkValue("field", registerDomain, requestScheme, fieldName).link();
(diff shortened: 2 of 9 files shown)


src/test/resources/test-app-config.yaml
```diff
@@ -8,6 +8,7 @@ server:
       port: 9001
 
 registerDomain: test.register.gov.uk
+registerScheme: http
```
michaelabenyohai, line 12: Should this also be added to config.yaml?

michaelabenyohai:
Happy to merge once we have changed the app config files.

src/test/resources/test-app-config.yaml
philandstuff, line 12: yes, good point

philandstuff:
closed in favour of openregister/openregister-java#9

PR #256, opened 2016-07-04
Implement Register proof resource

michaelabenyohai:
Using verifiable-log library from Bintray. Also using memoization.

### DIFF
--- a/src/main/java/uk/gov/register/presentation/RegisterProof.java
+++ b/src/main/java/uk/gov/register/presentation/RegisterProof.java
@@ -13,13 +13,13 @@ public class RegisterProof {
         this.rootHash = rootHash;
     }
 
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/register/presentation/RegisterProof.java
```diff
@@ -0,0 +1,27 @@
...
+public class RegisterProof {
+
+    private static final String proofIdentifier = "merkle:sha-256";
+    private final String rootHash;
+
+    public RegisterProof(String rootHash) {
+        this.rootHash = rootHash;
+    }
+
+    @SuppressWarnings("unused, used from template")
```
philandstuff: this comment isn't strictly true 😄

